Issue #9, opened 2019-01-11
readFolder doesn't return good urls


scenaristeur:
with this code 
```
  var url = this.url;
  console.log("fileclient reads",url)
  this.fileclient.readFolder( url ).then( folder => {
    if(!folder) console.log(this.fileclient.err)
    else console.log(`Read ${folder.name}, it has ${folder.files.length} files.`,folder);
    console.log("fichiers",folder.files)
    app.files = folder.files;
    app.subfolders = folder.folders;
    console.log("gestion des sous dossiers?",folder.folders)
    console.log("app.files",app.files)
    console.log("app.subfolders",app.subfolders)
  })
```
I've got this  : 
```
fileclient reads https://holacratie.solid.community/public/Tension holacratie-behavior.html.js:237:3
Read Tension, it has 13 files. 
[...]

PR #10, opened 2019-01-11
readFolder doesn't return good urls

scenaristeur:
https://github.com/jeff-zucker/solid-file-client/issues/9


jeff-zucker:
Many thanks!

PR #16, opened 2019-01-14
Update README.md

scenaristeur:
Couldn't find the requested release version 0.4.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ the two modules.
 Either download locally as shown below or use CDN like this:
 
 ```HTML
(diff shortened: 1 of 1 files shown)


jeff-zucker:
I really appreciate your helpful patches and sharp eye.  I went through a bunch of version numbers last night trying to get all the libraries in sync - I separated out sol into solid-shell and solid-auth-cli on its own now. I need to make sure that each of those interop with solid-file-client and vice versa.  To say nothing of solid-ide which I need to get back to someday.

jeff-zucker:
Oh dang, I just actually read your patch rather than skimming it.  You already got it out of the headers. Thanks,  I'll patch it now.

PR #40, opened 2019-05-11
Update README.md

scenaristeur:
some fix in examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Opens a popup window that prompts for an IDP then lets you login.
 
 ```javascript
 fileClient.popupLogin().then( webId => {
-    else console.log( `Logged in as ${webId}.`)
+     console.log( `Logged in as ${webId}.`)
 }, err => console.log(err) );
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -118,7 +118,7 @@ should prompt for a password.
 **logout()**
 
 ```javascript
-fileClient.logout().then( console.log( `Bye now!` )
+fileClient.logout().then( console.log( `Bye now!` ))
```
Otto-AA: If I'm not mistaken, it should be `fileClient.logout().then(() => console.log('Bye now!'))`.

Only passing `console.log('Bye now!')` passes the response of the console.log call as parameter, not a function which calls console.log

jeff-zucker:
Both these points are valid.  The login part of the readme also needs to be modifed to show it returns session from which session.webId may be derived.  @bourgeoa is now co-maintainer and may be able to get to this sooner than I will.  In any case, thanks @scenaristeur and @Otto-AA for your sharp eyes and suggestions.

jeff-zucker:
There was a typo in the example, if you downloaded before now, download again.  The correct flag to get the ACL files copied is fetcher.recursiveCopy( url, {copyACL:true})

jeff-zucker:
This is no longer relevant to v1.0.0. because we no longer provide session methods.

PR #43, opened 2019-06-06
Make interface more consistent

Otto-AA:
As discussed in #42 this PR aims to create a consistent response for the solid-file-client methods behavior while maintaining its simplicity.

## Src Folder
To achieve this, I've created following folder structure (for src):
- index.js                  // Handles the library exporting
- SolidFileClient.js    // Adds "easy" to use methods to SolidApi and use of solid-auth-cli(ent)
- SolidApi.js              // Class for making API calls to a solid pod (get, post, copy, ...)
- libraries.js              // For letting the user input rdflib
-> utils
- - apiUtils.js            // Some functions which are useful for working with APIs
- - folderUtils.js        // Functions for parsing solid pod folders

## Examples/Tests
In the examples folder, there is a browser example (examples/simple/index.html) which shows how to import the SolidFileClient, login via the popup and then get the contents of the profile card. There is also an example for node usage (examples/node/index.js) which imports and creates a new SolidFileClient. Both of these are not finished, but show how to use it in the different contexts. I've also created one test case in (tests/index.test.js) which fetches a public profile (it passes).

## Importing SolidFileClient
For now I've got importing via `<script src="dist/window/solid-file-client.bundle.js"></script>` and `const SolidFileClient = require('dist/node/solid-file-client.bundle.js')` working. In the test I've used `import SolidFileClient from 'src/index'`, but I don't know how exactly this is handled by jest. I didn't get script modules working in the browser yet (`<script type="module">import SolidFileClient from 'solid-file-client.bundle.js'</script>`).

## Build process
I've changed the build process a bit, so both builds for window and node are in the same config file now. I've removed some commands which seem redundant to me (cjs2em, "clean", cp src dist/node). I am not entirely sure if I understood the use case of the lib folder, but I think it is replaced by the src folder now (which uses imports and exports now).

## Dependencies
Regarding the dependencies, I've moved rdflib, solid-auth-cli and solid-auth-client to the peerDependencies for now, but I'm not sure if that is correct (ie if solid-auth-cli AND solid-auth-client are in there, does the user have to install both of them? If yes I would remove it from peerDependencies too). And there are some devDependencies which aren't used as far as I can see: mocha, chai, flow-bin, build-esm and eslint.
Regarding eslint, what do you think about using the "[standard](https://github.com/standard/standard)" style guide instead? It provides linting with a default config so we wouldn't need ten dependencies and configuration files for it (and `// disable xyz` in case we *really* want to do something). You can also pass it a `--fix` argument to make it automatically fix indentation, etc. And it is already used widely, so I think many javascript developers are used to that style.

## Simple API still missing
And obviously, the simple API is still missing. I've tried to create a good working environment first to make sure that we can build and test everything while developing. We still need to discuss how exactly we want them to look like before implementing them (also for the auth interface and the SolidApi if you have suggestions to make things differently). So that is left for future commits.

What do you think about these changes so far? If there is something you don't feel comfortable with, just say it. The changes I've made a more like a suggestion, not a definite end result.


jeff-zucker:
Wow, lots of work!  Thanks.  I'll go over all this more carefully before responding.

jeff-zucker:
I've tested, looks super!  Great job @Otto-AA!  

I'd like to keep the current master as master until we're ready to publish the new package.  So I've created a branch otto-changes.  Please submit the PR  to that branch.  

I think I like the way you handle dependencies.  Let the user bring their own everything.  I don't see a problem with  having them both in peer dependencies - it doesn't try to install them, just warns they're missing.

I am completely clueless when it come to lint so do whatever you think best.  Some of the dev-dependencies were added by linonetwo.

I understand using libraries.js for rdflib - it's cleaner not to have a global variable and we'd have to conditionally require rdflib in the nodejs version.  This way we don't need to mess with webpack, the user brings their own.  The part I don't like is the extra step of SolidFileClient.setRdfLib().  I would much rather pass both the auth and rdf in the constructor.  I tried importing libraries in SolidFileClient and setting libraries.rdflib in the constructor and it seems to work fine.  Any objection?  I was able to get readAndParse() working so rdflib behaves properly.

jeff-zucker:
Hmm, so the way you do it, the src uses import style and the nodejs version is only produced from it.  This is exactly backwards from my normal development flow.  I develop the src in nodejs and let babel create the import version in ./lib.  This lets me rapidly test and modify code from the command line without npm installing to create the bundle every time.  I do most of my testing in nodejs.   Apparently you have a different flow.  Is there a solution that works for both of us?

Otto-AA:
> I'd like to keep the current master as master until we're ready to publish the new package. So I've created a branch otto-changes. Please submit the PR to that branch.

Yeah, we should implement all desired methods, test everything and add proper documentation and examples before pushing to master. I don't really see what's the point of having it in solid-file-client/otto-changes instead of otto-aa/master until then. But if it would be useful for you I can do this.

> The part I don't like is the extra step of SolidFileClient.setRdfLib(). I would much rather pass both the auth and rdf in the constructor.

If we use externals, as @bourgeoa suggested, we would get rid of this and the libraries.js file. So if this works fine it would be a better solution from my point of view. Also this wouldn't require a conditional require as shown in #44

Otto-AA:
> This lets me rapidly test and modify code from the command line without npm installing to create the bundle every time. I do most of my testing in nodejs. Apparently you have a different flow. Is there a solution that works for both of us?

I usually try to write many automatic test cases which a cover a wide range of possible usages. Then I can run jest with the "--watch" parameter to make it run the tests every time I save some changes (`npx jest --watch`; I'll probably add a script in package.json for that). This way I see if any of the tests fails. After testing like this I usually make a build, load the script into the browser console and then work with it there.

If you prefer to require the solid-file-client into nodejs and then mess there with it you could try "npm run build -- --watch" (I'll likely add `npm run build:dev` as shortcut). This rebuilds the node version every time a file changes. For me all builds after the first take about one second, so I think that delay is okay to work with. This means if you run the build-and-watch command, testing would be the same as before except for a one second delay after saving. Would that workflow be okay for you?

jeff-zucker:
<blockquote>I don't really see what's the point of having it in solid-file-client/otto-changes instead of otto-aa/master until then. </blockquote>
Oh, duh, of course, that makes more sense.  I'll submit some PRs to you then.
<blockquote>npm run build -- --watch</blockquote>
OK, I guess I can try that way.

jeff-zucker:
<blockquote>If we use externals, as @bourgeoa suggested</blockquote>

If that works for you it works for me.  @bourgeoa, can you submit that as a PR on otto-master?

jeff-zucker:
@Otto-AA - are the methods in SolidAPI complete or just placeholders?  Specifically, are you intending to do any error trapping?  I am writing tests and it would be good to know how much trapping I need to do in the test itself.  Also, I'm going to start building the easy methods and need to know what to expect from the methods in SolidAPI I'll be calling with super().

jeff-zucker:
I see that you changed it so that this commit goes into jeff-zucker/otto-changes rather than keeping it in your repo.  So which way should we do it?  It seems like from here out @bourgeoa and I will be making more changes than you so it might be easier for you not to have to do the merging.  Makes no difference to me, but we should decide.

bourgeoa:
made PR #45 to replace `rdflib` libraries with webpack externals in `otto-changes`.
examples/simple is OK.

`npm install` never passed the test and for `npm run build` I moved `@babel/runtime` from peer to dev-dependencies. Is it a problem related to my configuration or do you have the same ?

jeff-zucker:
I have begun working on the fork in my repo so let's standardize on that.  Here;'s how to  clone it:
```
git clone -b otto-changes https://github.com/jeff-zucker/solid-file-client.git
```

jeff-zucker:
I have decided to go back to my original error-trapping before linonetwo refactored.  In each method I set the SolidFileClient.response to the full response of the method and return the body if any.  This means that methods should never fail and no user try/catch will ever be required for the easy methods.  SolidFileClient.response will always hold the response of the last method and we can do things like this:
```javscript
  let content = await fc.readFile(existingURL)
  if( fc.response.ok ) console.log("ok readFile on existing URL")
  else  fc.err("FAIL readFile on existing URL")

  let content = await fc.readFile(nonExistantURL)
  if( !fc.response.ok ) {
    if( fc.response.status == 404 )
      console.log("ok readFile on bad URL returns 404")
    else
      fc.err("FAIL readFile on non-existant URL")
  }
  else
    fc.err("FAIL readFile on non-existant URL")
```
SolidFileCLient.err() automatically appends the most recent response.status and response.statusText. 

What do you think?

jeff-zucker:
I am using the changes.log file to document the changes I am making at each step.  Y'all might want to do the same.  It will appear in my next commit in a few moments, also a working test.

jeff-zucker:
I am putting rdflib and solid-auth-cli as dev-dependencies.  This will mean that we can use them in tests and that users can have access to them with a simple npm install.  I am not sure if this means they should remain as peer-dependencies.  Any problems with this?

jeff-zucker:
Regarding testing against the bundle rather src : how do I deal with something like this?
```
l is not a function
    at s.readFolder (/var/www/html/solid/solid-file-client/dist/node/solid-file-client.bundle.js:1:4180)
```
All I have to do is find the 4,180th character of the one line script. :-)

Otto-AA:
> are the methods in SolidAPI complete or just placeholders? Specifically, are you intending to do any error trapping? I am writing tests and it would be good to know how much trapping I need to do in the test itself. Also, I'm going to start building the easy methods and need to know what to expect from the methods in SolidAPI I'll be calling with super().

They are pretty complete except for testing and error handling. I haven't tested all of them yet, but I will do that today. And if an error occurs (ie response.ok is false) for any api call it will throw this response in the future.

If you want to make a custom error for the simple API methods you could do as follows:
```javascript
readFolder(url) {
  const folder = await this.get(url)
    .catch(errorResponse => { /* Custom error handling */ })
  // Do something with folder...
}
// or
readFolder(url) {
  let folder
  try {
    folder = await this.get(url)
  }
  catch (errorResponse) { /* Custom error handling */ }
  // Do something with folder...
```

And you shouldn't need to call the methods with super(), calling them via `this.get(...)` should work too.

Otto-AA:
> npm install never passed the test

npm install shouldn't trigger the test imo. If we want we could make it build and test when installing (`prepublish: "npm run build && npm test"`), but I wouldn't do that. Changing `prepublish` to `prepublishOnly` should do the trick. I will make a PR with several other changes to package.json which includes this.

Otto-AA:
> I am putting rdflib and solid-auth-cli as dev-dependencies. This will mean that we can use them in tests and that users can have access to them with a simple npm install. I am not sure if this means they should remain as peer-dependencies. Any problems with this?

Seems good to me :+1:

Otto-AA:
You can pass set the mode of webpack to development (`npm run build -- --mode development --watch`). Then the code doesn't get compressed. I will add a `npm run build:dev` script which is a shortcut for this.

But why would you want to test against the build rather than the source? To verify that nothing gets broken in the build process?

jeff-zucker:
OK, lol, since you weren't throwing errors in the advanced, I assumed you were going to always return a fetch response with ok, status, etc. and I changed all my code to do the same.  Are you sure that's not what we want to do, it is actually very clean that way.  I'll change mine back to reject if that makes more sense.

jeff-zucker:
<blockquote>why would you want to test against the build rather than the source</blockquote>

Because the source is in import style and doesn't run with node.  It is so much more lightweight to be able to just code/test/code/test without babel and jest.  But I'm getting more used to doing it this way so I'll just stop complaining about it. :-)  The mode tip will help, thanks.

jeff-zucker:
I feel that createFolder() should definitely fail if the parent folder doesn't exit.  That's what the spec says and that's how NSS operates.  Put should create intermediate containers, but not post.

jeff-zucker:
Oh wait a minute, you mean throw an error if the folder itself is found. That is bad in my opinion.  The way Solid works, if the folder exists and you try to create it, it will create a duplicate with a random number attached which will be returned in the location header.  This behavior is sometimes desirable so it should be supported somewhere.  OTOH, the new user would not expect this and so there should be a way to ask to create a folder but only if it doesn't exist.  In that later case, 404 is not really failing, it is succeeding.  So it looks to me like we need two methods - the easy one would succeed with 200 if the folder pre-exists and with 201 if it doesn't; the advanced one would carry out the creation of the folder regardless of whether it pre-exists or not.

jeff-zucker:
If you agree that we need both ways of handling createFolder, perhaps createFolder could be the easy one and makeFolder the advanced one?

jeff-zucker:
Your import of $rdf in folderUtils does not work in a nodejs environment. Obviously my klduge of passing the variable was not optimal but at least it worked.  Perhaps we can make folderUtils part of the class and let it get the rdflib on initialization.  We need to do everything the same way: we can't have a "user brings their own" policy and then go importing it in the module.  The only way I could see your import working is if the rdflib is installed globally and we can't assume that is the case.

jeff-zucker:
OK, I was wrong that rdflib in folderUtils didn't work in node, it does.  But it doesn't get picked up in the jest environment, I get "can't find method graph of undefined"  Whereas same calle to readFolder() works fine outside of jest.

jeff-zucker:
I finally got the rdflib thing taken care of, sorry about the crappy refactor to pass it from one to the next.  The problem turned out to be a differential between what require and import want as pathnames.  Finally got the right combination which works in both commonJS and ES6 .

Otto-AA:
> OK, lol, since you weren't throwing errors in the advanced, I assumed you were going to always return a fetch response with ok, status, etc. and I changed all my code to do the same. Are you sure that's not what we want to do, it is actually very clean that way. I'll change mine back to reject if that makes more sense.

Sorry for being unclear about that in the beginning. I think the solution in #48 is a nice solution to this. The throw-catch way seems cleaner to me because no code accidentally gets executed with failed requests if someone forgets response.ok and also you don't always have to check for response.ok (e.g. in the recursive copy-method I'd need to check for response.ok for every fetch. with throwing the error just propagates upwards). But I think both solutions have their benefits so #48 seems nice :+1:

Otto-AA:
> If you agree that we need both ways of handling createFolder, perhaps we could still have one method but with a "allowDuplicates" flag.

I could add a `ensureFolderExists` method to SolidApi which uses a PUT request if the folder is not existing already. And `createFolder` uses POST to create a new folder (if there is a parent). Or do you think that's too confusing?

Otto-AA:
> I finally got the rdflib thing taken care of, sorry about the crappy refactor to pass it from one to the next. The problem turned out to be a differential between what require and import want as pathnames. Finally got the right combination which works in both commonJS and ES6 .

Sorry for troubling you with this. I thought that the commit which removed passing rdflib worked in node too. But I'm glad that it works now!

PR #44, opened 2019-06-06
rdflib and solid-auth-client as externals

bourgeoa:


jeff-zucker:
This was done in #45, so I can close this, right?

PR #47, opened 2019-06-08
Update tests, rdflib import and error handling

Otto-AA:
* separated test folder into test-folder/solid-api and test-folder/easy
    to make sure they don't interfere with each other
    test-folder automatically gets cleared before each test run (see setup.js)
  * folderUtils now imports rdflib instead of passing it around
  * all "advanced" API methods except for fetch now throw if !response.ok
    to match this behavior, createFolder now rejects if parent doesn't exists
    (this will probably be discussed and changed in the future)
  * fixed deleteFolderContents and added one test for it

Do you think that `createFolder` and `readFolder` should be acting like simple or advanced methods? From the advanced users point of view they are essential methods and therefore it would be nice to have them reject on a failure. From the simple usage point of view, they are also essential and therefore should behave like other simple methods. I don't think that naming the simple and advanced methods would make much sense, as their job is exactly the same. Hence I'm not really sure what we want to do with methods like these.

Could you maybe explain the reason a bit more, on why you want to resolve errors for the simple methods? As far as I've seen the solid-file-client in the master branch also rejects on errors, so it would be a change to the current version.


jeff-zucker:
As  I said in the other thread, I assumed since you weren't throwing errors that I shouldn't :-).  But now that  I've done it, I find it very clean.  The user doesn't need to catch anything, always has access to the full response,can easily test for .ok and .status.   What are your reasons for want to throw errors?

PR #50, opened 2019-06-11
Set SolidApi.fetch to throw and added new tests

Otto-AA:
* to be consistent fetch now throws if response.ok is set to false
    I've updated methods in SolidFileClient to use get and handle thrown errors
  * added some tests for SolidApi
  * added tests for apiUtils 

In the tests I've encountered something unexpected regarding the post method: If I try to create a folder which already exists with post it responds with 409 instead of creating a new one. And trying to overwrite a file resolves with 201 but I don't see a new one being created. Is this something related to solid-auth-cli?

### DIFF
--- a/changes.log
+++ b/changes.log
@@ -61,4 +61,8 @@ JZ 2019-06-09 22:00 PDT
   * the addresses can then be used to delete,copy,etc. the .acl and .meta files
     using normal methods
 
-
+OA 2019-06-11 03:00 PDT
+  * to be consistent fetch now throws if response.ok is set to false
+    I've updated methods in SolidFileClient to use get and handle thrown errors
+  * added some tests for SolidApi
+  * added tests for apiUtils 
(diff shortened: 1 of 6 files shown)


jeff-zucker:
It looks like your changes to SolidFileClient are on a version many commits ago.  I do not know what the impact of these will be if I merge.  Please separate out the commits so that one PR is for changes to your packages and if you have changes for SolidFileClient put those in a PR of their own after doing a pull of my most recent commits.

jeff-zucker:
I'll look into the file:// problems.  You can always switch to using remote files to see if this is a problem in solid-rest-file.

jeff-zucker:
I guess I'll just backup and merge and assume that the changes you made to SolidFileClient will not clobber things I've done since.

Otto-AA:
I didn't notice that I was behind on my local branch, so I've merged your changes into mine afterwards via github (see https://github.com/jeff-zucker/solid-file-client/pull/50/commits/86a685f4d4009cd59d0b220bd249eef0e82d926f). While this only was a little change to do, it also added your changes so that's why many things seem duplicated. I'll try to check and git pull from now on before commiting.

jeff-zucker:
Not a problem, I merged any everything looked good.

PR #51, opened 2019-06-12
Run tests from src rather than dist

Otto-AA:


jeff-zucker:
Any idea why switching from dist to src dropped SolidFileClient.js coverage from 84% to 21%?

Also, is there any way to get the coverage to not display some of the extraneous stuff like the externals?

jeff-zucker:
Nevermind about the coverage thing - I had to change jest.config.js and session.test.js - the first was ignoring src and the second was using it so 20% was what session  covers.  With the  fix to jest.config.js, it's back to 80%+

Otto-AA:
Oh, that's likely because of the session test.

You should be able to remove the coveragePathIgnorePatterns now, because node_modules is the default anyway. If you want you can set `collectCoverageFrom: [ "./src/**" ]` to be exact on what we want to be included, but it shouldn't make a difference for now.

jeff-zucker:
On mock, I may have a different solution (in the long run) - if solid-rest and its descendants respond to login requests with mock data, we can test against logging in to file:// or app:// and that way, only the base URL needs to be changed to run the test as real against a pod or mock against solid-rest mini-pod.

jeff-zucker:
Because, obviously, we should run all tests against a real pod manually before release.

BTW are you still experiencing things that might be broken in solid-auth-cli/solid-rest-file?

Otto-AA:
> BTW are you still experiencing things that might be broken in solid-auth-cli/solid-rest-file?

Not really. I will add tests for the SolidApi which will use solid-auth-cli so I will probably encounter unexpected behavior. But I think that the tests for the core methods (fetch, get, head, ...) should already cover that and that copy and others won't notice any more errors in solid-auth-cli.

PR #54, opened 2019-06-13
Add TestFolderGenerator and update SolidApi tests

Otto-AA:
I've added a new class which I'm using to generate folder structures (see example below).

The parsing still doesn't seem to work @jeff-zucker . If you comment out the deleteFolderContents test in line L117 in SolidApi.test.js you should see the error produced with the new version. 

```javascript
/** Example generating a testing folder structure
 * The generated structure is:
- SolidApi
- - my-test
- - - child
- - - - file1.ttl
- - - file2.ttl
 */

const base = "file://" + process.cwd() + "/test-folder/"
const container = new RootFolder(base, "SolidApi")
const fileOne = new File('file1.ttl', content)
const myTest = new RootFolder(container, 'my-test', [
  new Folder('child', [
    fileOne,
  ],
  new File('file2.ttl')
])

await myTest.reset() // Deletes my-test folder and then creates the folder structure

console.log('url for testing', myTest.url)
console.log('fileOne url', fileOne.url)
```
(Not tested, if you want a tested example I can write one later)
RootFolder creates a folder which already has a base path set. It will set the base path for all children automatically. Hence the `new Folder`s only need the name. If you want to use the url of a created file/folder, you'll need to store it as a variable and then pass it as parameter to a RootFolder (passing it to multiple root folders could break its base path, so only use it once). If you want to test out if the path works, you can pass it the `dryRun: true` option. 
If you think this is useful for you and want something better explained, just ask. I'll use it for my tests because it makes it easy to create sample directory structures for testing recursive methods. It should be able to implement it with a local storage version of solid-rest if desired (I'd prefer that because messing with the file system is always a bit risky, so I'll try to implement that at some time).

### DIFF
--- a/tests/SolidApi.test.js
+++ b/tests/SolidApi.test.js
@@ -1,39 +1,29 @@
 import auth from 'solid-auth-cli';
 import SolidApi from '../src/SolidApi'
 import apiUtils from '../src/utils/apiUtils'
+import { Folder, File, RootFolder } from './utils/TestFolderGenerator'
 
 const { LINK } = apiUtils
 
-const base = "file://" + process.cwd()
-const folder = base + "/test-folder/SolidApi/"
-const inexistentFolder = folder + "inexistent/"
-const inexistentFile = folder + "inexistent.abc"
-const turtleFile = folder + "turtle.ttl"
-const turtleFileContents = "<> a <#test>."
+const base = "file://" + process.cwd() + "/test-folder/"
+const container = new RootFolder(base, "SolidApi")
+const inexistentFolder = container.url + "inexistent/"
+const inexistentFile = container.url + "inexistent.abc"
 
 const api = new SolidApi(auth.fetch.bind(auth));
 
-// Create container
-beforeAll(() => api.createFolder(folder))
-beforeAll(() => api.createFile(turtleFile, turtleFileContents, 'text/turtle'))
+// Reset container
+beforeAll(() => container.reset())
 
 describe('core methods', () => {
-  const coreFolder = folder + "core/"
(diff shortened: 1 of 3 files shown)


jeff-zucker:
While I understand you wanting to use rimraf because of the problems created with the earlier method, I strongly recommend that you do not.  We need to run all of these tests in the browser and against NSS, otherwise they are meaningless.  We don't need to necessarily automate that part - we might have a test suite as we do now and then some tests that we would run manually requiring login, etc.   But the less we assume about the environment of the tests the better.  Rimraf works only with solid-rest-file, not with any other solid-rest storage handler or with https.  OTOH, I am fine with whatever ad hoc tests you want to do.  I am a fan of rm -rf test-folder; npm test, but it's limited.  I am in the middle of writing a scheme-neutral testing initializer.  Will let you know as it proceeds

I am testing PUT, POST, and DELETE - I want to make triple sure that solid-rest-file outputs the same things as NSS.

Otto-AA:
The TestFolderGenerator class could also be rewritten to use https or any other other storage instead of fs. I think writing a test package which works with only the fetch method provided by solid-auth-client/solid-rest could be very useful too. With such a test package one could easily change between https, fs, in memory storage, etc. But I'll wait until you've finished your testing initializer before thinking more about this.

jeff-zucker:
Ha lol.  I've already done that.  Test all the rest methods against file, https, and app all in one test.  I am 15/15 on file and https and about 5 fails in app.   The most surprising thing about the https test is that the server sends 400 if you put a slash in your POST slug!

PR #57, opened 2019-06-19
use of  null-loader in webpack for solid-auth-cli

bourgeoa:
build OK, test with solid-ide OK

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "solid-file-client",
-  "version": "0.5.1",
+  "version": "0.5.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


theWebalyst:
@jeff-zucker is this something that should also be applied to solid-auth-client? In fact anywhere that is using `externals { something: 'null' }`

bourgeoa:
I would say yes, whenever you got the no null module message in compile. If you use browderify/babelify there is an other solution with adding an  ignore null module parameter.

bourgeoa:
I would say yes, whenever you got the no null module message in compile. If you use browderify/babelify there is an other solution with adding an  ignore null module parameter.

jeff-zucker:
Mark, @bourgeoa knows more about this than I do.  But the line in rdflib's webpack that defines solid-auth-cli as null is to avoid putting it in the bundle even though it IS a dependency because it is never used on the web.  That is a bit different from using null to define externals which, as I understand it, are items that are  NOT dependencies but the user should bring to the game e.g. with a script tag. It is this second usage which we are discussing for solid-file-client so that the package needs to have solid-auth-client or solid-auth-cli available but does not want them in the bundle.

theWebalyst:
Thanks @jeff-zucker, the mists are gradually clearing in this area... only for them to form again in a week or two unfortunately! 😉

PR #58, opened 2019-06-20
Update test environment

Otto-AA:
The main change is the new tests/utils/contextSetup.js file which looks up the testing environment and provides a few methods to (1) set up the testing environment and (2) getters for the baseUrl, prefix, fetch, auth and testContainer. I am not 100% happy with how it is done, but I think it is good enough for this project. The https:// prefix isn't working yet, app://ls/ and file:// are passing the tests.

# General idea
There are two environment variables used as config:
`process.env.TEST_PREFIX`: either "app://ls/", "file://" or "https://"
`process.env.TEST_BASE_URL`: (only used by https) selects in which container the test-folder will be created, e.g. https://test.solid.community/private/tests/
Depending on how these are set, the contexSetup.js script will select and initialize a specific testing environment. This can be used by other tests. 

# context setup
`contextSetup()`: This method sets up the baseUrl, testFetch and the auth
baseUrl: That's where the tests will be run. The folder it points to is required to already exist.
testFetch: asserts that no requests are made which don't start with the baseUrl (ie accidentally delete root). If the url is ok it just forwards to solid-auth-cli. Could also be used to log requests.
auth: An instance of solid-auth-cli with the fetch method overwritten to the testFetch. If https:// is used it will be logged in

# Things to consider when using
The getters (e.g. `getFetch()` or `getBaseUrl()`) will throw if `contextSetup()` hasn't been finished yet because they can't know in which context they would run. As a result, one cannot get the url of a TestFolderGenerator item (e.g. BaseFolder(...)) before contextSetup has run, which in jest means container.url/container.reset will need to be used within `beforeAll/beforeEach/test/...` but not directly within `describe`. To work around this issue I've introduced FolderPlaceholder and FilePlaceholder, see SolidApi.core.test.js for an example.

To use an environment, just call `TEST_PREFIX=app://ls/ npm run test` or use the scripts I've added to the package `npm run test:ls`.

# Prefix https:// not working yet
As mentioned above, the https:// prefix isn't working yet, but I couldn't find the error so I've opened this PR anyway now. I've called it with:
```sh
SOLID_IDP=https://solid.community/ SOLID_USERNAME=otman SOLID_PASSWORD=********* TEST_PREFIX=https:// TEST_BASE_URL=https://otman.solid.community/private/tests/ npx jest tests/testSetup.test.js
```
The setup.js script works fine and creates the https://otman.solid.community/private/tests/test-folder/ folder. But strangely the server responds with 401 for the requests from testSetup.js (which uses TestFolderGenerator which uses the testFetch and the logged in solid-auth-cli). The server also sets the header WWW-Authenticate to `Bearer realm="https://solid.community", scope="openid webid", error="invalid_token", error_description="Invalid PoP token signature"`. I've checked that the Bearer is set (it differs to the one used in setup.js but looks ok). So I don't know what else could cause this trouble. Do you have any idea what could cause this?
(And if you want to debug it, I'e added an if (false) { console.log...) part to SolidApi.createFolder and contextSetup.testFetch which you can set to true to see more logs)


bourgeoa:
You are doing the tests in  `solid.community/private/` 
Have you tried `public/`. I remember private having a weird .acl so as to be used mostly for notifications. Even for the owner.

On top of that when I go to my https://bourgeoa.solid.community there is no default `private/`. On Osman pod I have no permission to private/

So I would check .acl for private/

Otto-AA:
Thanks for your suggestions, I've tried it with public now but it also doesn't work (also creates the test-folder directory in setup.js, but authentication fails for testSetup.test.js afterwards).

I think if it was because of an acl restriction it would respond with 403 Forbidden instead of 401 Unauthorized + the "Invalid PoP token signature".

jeff-zucker:
I thought you had said you were not going to be using SolidApi in setup because that would require it to work before it was tested.  But I see all your calls are to it.   Change of mind?

jeff-zucker:
I tried to find the cause of the 401 but I got lost in all the times you pass the various fetches around.  Part of the problem is that SolidApi gets only the fetch method not auth itself so we can't do auth.currentSession to see if we're logged in inside SolidApi.  Is there a reason you don't want the entire auth object there?

Otto-AA:
> I thought you had said you were not going to be using SolidApi in setup because that would require it to work before it was tested. But I see all your calls are to it. Change of mind?

I am still not really happy with it, but it seems like the best way to test in app://ls/ and https://. If I have time I will add a fs option which uses mkdir, rmrif and stuff in the TestFolderGenerator instead so we also have an independent test setup.

> I tried to find the cause of the 401 but I got lost in all the times you pass the various fetches around. Part of the problem is that SolidApi gets only the fetch method not auth itself so we can't do auth.currentSession to see if we're logged in inside SolidApi. Is there a reason you don't want the entire auth object there?

I will create a version which uses auth instead. I am pretty confident that I'm logged in (because it passes a authentication token to the request which returns 401), but maybe it makes it easier to debug.

Otto-AA:
Oh, and the `response.status !== 404 && response.status !== 401` in SolidApi.createFolder should only be `response.status !== 404`. Forgot to remove this after trying to find the error

Otto-AA:
Here's the version which uses auth and doesn't overwrite the fetch method of solid-auth-cli: https://github.com/Otto-AA/solid-file-client/tree/test-environment

As far as I've seen currentSession returns good results (right before attempting to delete/create items).

Otto-AA:
I've narrowed the error down, but got even more confused. I've got the same code running in the jest setup and in a beforeAll() method running, but in the setup it works while it fails in the beforeAll method. If you have any ideas, I would highly appreciate them :)

Here is the code I am using in both files (available in the test-environment branch linked above):
```javascript
async function setup(name) {
  console.group(`setup[${name}]`)
  await auth.login()
  const url = process.env.TEST_BASE_URL + name + '/'
  const fileClient = new SolidFileClient(auth, { throwErrors: true })
  const res = await fileClient.createFolder(url)
  console.log(`finished setup[${name}] with ${res.status}`)
  console.groupEnd()
}
```

To run, enter following with your username, password and baseurl:
```sh
SOLID_IDP=https://solid.community/ SOLID_USERNAME=otman SOLID_PASSWORD=******** TEST_PREFIX=https:// TEST_BASE_URL=https://otman.solid.community/public/tests/ npx jest
```
Before running ensure that the tests/ folder exists and is empty. When running this test, a "setup" folder will be created and a "beforeAll" folder should (!) be created too.

It also works if I use the node build, so I am pretty confident that it is something related to jest. I will take a look at how high-level-crud-response.test.js got it working and what the difference between them is.

bourgeoa:
@Otto-AA I found a solution to your https test error.
comment out : 

> // await auth.login()

and you pass the test. Could be you cannot login more than once, and you are allready logged in

Otto-AA:
Thanks for the tip. I think it could be useful as a workaround for now, but I fear it is not fixing the error itself.

If I remember correctly you have set your acl permissions to allow anyone to read and write, so if you comment out "await auth.login" the fetch won't send authentication credentials but it will still work because they aren't required. I've tested that with my public folder, but I get a permission error this way (with the default acl for public on solid.community). I've also tried to only login if no currentSession is found, but that resulted in the same error.

We could use no authentication and test folders which can be written by everyone in the meantime though.

bourgeoa:
It seems I allways had the problem in using solid-file-client in my project tiddlywiki-node-solid-server
Could it be because the fetch used is not allways the solid.auth.fetch ?

bourgeoa:
I just think to the bug you declared for v5.1.5 and the discussion following about a bug in POST.

Otto-AA:
>  Could it be because the fetch used is not allways the solid.auth.fetch ?

I am using solid-auth-cli, so I don't think that solid.auth.fetch should be used at all. And I've checked that solid-auth-cli uses ifetch with the authentication tokens, that works fine.

> I just think to the bug you declared for v5.1.5 and the discussion following about a bug in POST.

This shouldn't affect solid.community afaik, as this still uses an older version.

jeff-zucker:
If you do this you are, which is what I think @bourgeoa meant :-)
```
let solid = { auth: require('solid-auth-cli') }
```
Which is how I usually call it in scripts I intend to call in both browser and node.  Other than the require line ,it's the same code for either.

jeff-zucker:
I took a break from coding this afternoon to participate in this.  I thought Otto might appreciate it.
IMG_0175

jeff-zucker:
My first climate change action was in 1994 when I created a website and document creation system for the International Institute for International Development who did live reporting at  at all the UN climate summits.  I worked for them and a number of other climate change organizations over the years including a decade working at the UN on that and other issues.

Otto-AA:
> If you do this you are using solid.auth.fetch, which is what I think @bourgeoa meant :-)

Ok, thanks for the explanation

> I took a break from coding this afternoon to participate in this. I thought Otto might appreciate it.

It is good to see people walking up for a better environment protection all over the world :)

Otto-AA:
@jeff-zucker Does the high-level-crud-throw.test.js works with https? I've tried it, but also get the 401 response. Did it work for you at one point, or does it still work?

jeff-zucker:
Sorry, I'm in the thick of trying to refactor processFolder and don't have
a working package to test.  Will test as soon as I do.

Something to keep in mind: solid-auth-cli stores the session in a global
variable so two different solid-auth-cli objects will not share a session.
At some point I may try to store the session to disk, but unless you can't
fix things the way they are, I'd much rather not try to do that while
working on solid-file-client.  Also solid-auth-cli is currently not a
class, just a bunch of functions which I originally did to mimic
solid-auth-client.  Is there a way I can turn it into a class but keep the
functional interface (i.e. the user doesn't need to call new)?

Otto-AA:
> Sorry, I'm in the thick of trying to refactor processFolder and don't have a working package to test. Will test as soon as I do.

You could git stash the changes, git checkout an old commit and test it there. When finished checkout the branch where you left and then apply the git stash again (I don't know the exact commands, as I'm using my IDE for stash). I understand if you don't have the time to do that, just wanted to note it in case you need it in the future :)

> Something to keep in mind: solid-auth-cli stores the session in a global variable so two different solid-auth-cli objects will not share a session. At some point I may try to store the session to disk, but unless you can't fix things the way they are, I'd much rather not try to do that while working on solid-file-client. Also solid-auth-cli is currently not a class, just a bunch of functions which I originally did to mimic solid-auth-client.

I've also tried to use solid-auth-cli from one file only, so there is no possible interacting with other instances (as in the [test-environment](https://github.com/Otto-AA/solid-file-client/tree/test-environment) branch, but with the globalSetup jest config set to undefined (?)). So this shouldn't be the problem.

> Is there a way I can turn it into a class but keep the functional interface (i.e. the user doesn't need to call new)?

Yeah, I've tried something out and it worked: https://jsfiddle.net/bperx8gs/
I think it could break things if you use names like "prototype", "bind", etc., but I think if you don't use such names it should be ok. I am considering to use a similar approach for the throwing vs resolving responses option in this project, but I think we should leave that to the end.

jeff-zucker:
Thanks for the stash info - my usual way is copy the current to a backup,
pull the thing, replace it with parts of the backup if necessary, very
manual and kldugy.

And thanks for the function -> class thing.  I'm not sure if I explained
what I need correctly or if I did, Im not sure how to use your example.

Currently user does const auth = require('solid-auth-cli); auth.login()
...  I do not want that to change or have the user have to first require an
object and then call new on it.  But I do want solid-auth-cli itself to be
a class and export something that allows the old invocation to work.  Does
that make any sense?

Otto-AA:
> Currently user does const auth = require('solid-auth-cli); auth.login() ... I do not want that to change or have the user have to first require an object and then call new on it. But I do want solid-auth-cli itself to be a class and export something that allows the old invocation to work. Does that make any sense?

Yes, that's what I try to do in the example I've made. The A class should contain all methods solid-auth-cli currently contains (fetch, currentSession, ...). This class could be used to create multiple instances of solid-auth-cli.
For backwards compatibility it creates a default instance and kind of merges the class and the default instance (it adds the methods of the default instance to A). This merged object could be used with `new A(...)`, but also is an instance of this class and therefore has .fetch, .currentSession and co. So if you export it, it should be backwards compatible and be usable with "new" :)

In your case, class A would be class SolidAuthCli (line 1). You create a (default) instance of this class for backwards compatibility (line 13). Then merge the default instance with SolidAuthCli (line 16). Then export it. It can be used as "SolidAuthCli.login" and "new SolidAuthCli" now. (In my example I tried to show that, but it should be `A.log()` in line 23 which shows that it works).

Note that I've never done something like this and don't know if that is good to do or has bad drawbacks.

jeff-zucker:
Thanks much for the explanation of the class situation.  I'll experiment
when I have time (ha!) and let you know.

jeff-zucker:
I just committed a refactored processFolder.  It's still a work in progress but what I did was completely refactor to separate the ugly bits out to folderUtils and split both processFolder and getLinks into several smaller methods. Not done with the refactoring, but it's a start. I also implemented the "withLinks" flag with false being the default.  That now works, and fixes some bugs in the previous version.  The methods for the withLinks = true are not ready yet. 

I would like to return the folder object itself from readFolder, not stash it in a body.  But I didn't want to change that yet because it would impact your code.

I ran the setup and it failed at createFolder with 401.  I would suggest trying to define auth at the very top of the chain (setup.js?) and pass it to whatever other things need it.  I'm guessing it is a jest thing, because I don't have that issue with my non-jest scripts.  But, yes, sadly I get 401 when I run in jest using your env variables and https.

Otto-AA:
> I would like to return the folder object itself from readFolder, not stash it in a body. But I didn't want to change that yet because it would impact your code.

I also think that would be a good idea. There are only two lines in which I use readFolder ([273](https://github.com/jeff-zucker/solid-file-client/blob/otto-changes/src/SolidApi.js) and [308](https://github.com/jeff-zucker/solid-file-client/blob/otto-changes/src/SolidApi.js#L308)). Feel free to change the readFolder return value and these two lines.

Otto-AA:
> I ran the setup and it failed at createFolder with 401.

Which script exactly failed for you now? high-level-crud-throw.tests.js, setup.js or testSetup.test.js?

> I would suggest trying to define auth at the very top of the chain (setup.js?) and pass it to whatever other things need it. I'm guessing it is a jest thing, because I don't have that issue with my non-jest scripts.

The problem with this is, that we can't share an object between jest tests afaik, only json-serializable data. I'll try to store the serialized session in setup.js and load it into the tests, but I fear that this will result in the same error. If that doesn't work either, I'll try copying the authentication token which triggers 401 and make a manual fetch to see if I get the same response with this token (ie if the token is invalid or something in the fetch process fails). But I don't think that I'll dig much deeper. We'll likely need to go with the everything-allowed acl suggested by @bourgeoa and no authentication.

jeff-zucker:
The problem occurs in setup at createFolder.  I do not get a 401 in my
tests because I login within each test.  That would probably solve it for
you as well.

Otto-AA:
Sorry, I am not sure if I understand what you mean. Does the problem occur in setup.js or in the setup of testSetup.test.js (of [this]() branch)?

And I login once per test file in the beforeAll() clause. Is this what you mean, or do you really mean for each test()? I haven't tried logging in directly before test(), but logging in in beforeAll() also fails for me.

In other words, what would you change in testSetup.test.js (below) to make it work?
```javascript
import SolidFileClient from '../src/SolidFileClient'
import auth from 'solid-auth-cli'

async function setup(name) {
  console.group(`setup[${name}]`)
  await auth.login()
  const url = process.env.TEST_BASE_URL + name + '/'
  const fileClient = new SolidFileClient(auth, { throwErrors: true })
  const res = await fileClient.createFolder(url)
  console.log(`finished setup[${name}] with ${res.status}`)
  console.groupEnd()
}

beforeAll(() => setup('beforeAll'))

test.only('ignore me', () => {})
```

jeff-zucker:
The tests fail with 401 at tests/utils/TestFolderGenerator.js:105

My suggestion was to login in once in every test suite, not every test.

Otto-AA:
> The tests fail with 401 at tests/utils/TestFolderGenerator.js:105

I've meant for the test-environment branch, but I suspect testSetup.test.js then.

> My suggestion was to login in once in every test suite, not every test.

This is exactly what I'm doing in the test-environment branch, but it fails with the same error (401 with invalid poptoken). You can either clone the test-environment branch (`git clone -b test-environment https://github.com/otto-aa/solid-file-client`) or just copy paste the code above and run it with jest (`SOLID_PASSWORD=*** ... npx jest tests/testSetup.test.js`). It will very likely fail at createFolder with 401.

Otto-AA:
Would it be possible that you show me the code which works for you in jest with https? I am still struggling with getting it to work, the only possible way I see until now is to use a test folder where everyone has access to. So an example which I could compare to the tests which fail for me would be nice to have :)

jeff-zucker:
I'm sorry if I've been unclear but I haven't touched jest since the setup
has been broken other than to test it when you requested.  i have a lot of
work to do to make sure readFolders and rdfQuery and localStorage all work
(all of those are things I've only released since we've been working on
solid-file-client and they need to work and I need to test them.  So I test
in node.  I'll be committing the tests soon, but they won't help with jest
issues.

PR #60, opened 2019-06-21
Add more tests and standard linting

Otto-AA:
I've added a few more tests for the composed methods and also added the [standard](https://github.com/standard/standard) linting dependency.

@jeff-zucker There are some lines I wasn't sure how I should change them while fixing the linting errors:
(execute `npm run lint` to see the exact line numbers)
In SolidApi:
the scope of aclres is restricted to the try block, hence it isn't visible afterwards.
we shouldn't use alert for errors
the standard linter says that there are some unnecessary escaping slashes in some regex expressions (I guess it is right, but didn't want to just change it)

In SolidFileClient:
The linter doesn't understand the "do" method. I guess it is a valid name, but I didn't see it being used anywhere. I guess we can remove it for now as we also always throw now?


Otto-AA:
The new commands I've added:
`npm run lint`: Checks for linting errors (indentation, spaces, unused variables, etc.) and outputs them to the console
`npm run lint:fix`: Tries to fix all linting errors and outputs those where it didn't know how fix them

Otto-AA:
And the tests currently fail due to the missing import of the upcoming rdf package. I've tested it prior to merging and they were fine :)

jeff-zucker:
> And the tests currently fail due to the missing import of the upcoming
rdf package

I have no idea what you mean.  What package are you talking about?
src/utils/rdf-query.js is the only RDF package we will be using.

jeff-zucker:
I am going to move all of uriJoin and most of parseLinkHeader to folderUtils.  The portions used there will be the portions that do not have any rdf code so they will do no imports.  The small section of parseLinkHeader will go in a separate method that should remain with getLinks and processFolder because the three are inter-related.

jeff-zucker:
BTW, I note that v0.5.1's browser bundle is 547k and now, after switching to N3, v1.0.0 is 112k.

Otto-AA:
> I have no idea what you mean. What package are you talking about? src/utils/rdf-query.js is the only RDF package we will be using.

Yes I mean that the import of src/utils/rdf-query.js fails, therefore SolidApi and basically everything using SolidApi doesn't work currently.

> I am going to move all of uriJoin and most of parseLinkHeader to folderUtils. The portions used there will be the portions that do not have any rdf code so they will do no imports. The small section of parseLinkHeader will go in a separate method that should remain with getLinks and processFolder because the three are inter-related.

Sounds good

> BTW, I note that v0.5.1's browser bundle is 547k and now, after switching to N3, v1.0.0 is 112k.

Yeah, rdflib is pretty huge, so getting rid of it makes it more light weight. For a direct comparison we would have to calculate solid-auth-client to it which is itself 124k, so we've got ~550k previously to ~240k now.

jeff-zucker:
> Yes I mean that the import of src/utils/rdf-query.js fails, therefore SolidApi and basically everything using SolidApi doesn't work currently.

That import works consistently for me.  But I haven't been testing in jest.  In the future, if I commit something that doesn't work, please let me know.

I am doing a massive refactor or processFolder and getLinks, I think you will like the results much better.

Otto-AA:
Maybe this works for you, but in the version you've committed src/utils/rdf-query.js is only a path (`/home/jeff/Web/solid/rdf-light/src/index.js`) to the file, not the file itself. That's why I called it black box and "One line of code for the RdfQuery class?" in the other issue :)

https://github.com/jeff-zucker/solid-file-client/commit/83dea8b8ffb3231b3ee6d9ed3492a0b8bac4379c#diff-550d64b9f0a45057366e0c0699bb76fd

jeff-zucker:
Oh, well it's pretty hard to guess that you mean there was no file there from descriptions like "a black box" - I thought you were referring to my coding style.  When you said "one line of code, I thought you meant that the query itself did in one line what you need a fetcher and named nodes and other things to do with rdflib.  Now that you have actually told me that there is no file there I see that I have a symlink and git status told me it was uploading it, so I had no reason to suspect it wasn't uploading it.  I'll remove the symlink and commit.

Otto-AA:
Ok, I somehow thought you used it as a placeholder for the npm package. I'll be clearer with my comments next time :)

jeff-zucker:
Phew, ok, now it is there, I went to the repo and checked.  I'm glad there's an explanataion :-)

Otto-AA:
Ok, thanks for committing the file.

I've still got issues with the test though:
In SolidFileClient there's a syntax error (one closing bracket too much on line 166)

The error handling of processFolder and the rdf file doesn't seem good enough for me:
First of all, the loadFromUrl method should check whether or not the request was successful. This caused an undefined error later on. And if loadFromUrl throws it should be catched and handled properly (dunno if this is the case. I'll just mention it so you can check)
In processFolder, I've removed the check for files.length === 0, because this resulted in the files and folders properties being undefined (rather than [] and [folder]). 

The few tests of mine which rely on readFolder now work, but some of yours still don't. I don't have the time to check them out currently, so I've pushed the few fixes to #61

PR #65, opened 2019-06-26
crud-browser.html

bourgeoa:
crud-browser.html with popupLogin() from solid-auth-client 

Some other changes are indicated on end line with changes (hope no other changes)
69, 145, 160

all 24 tests passed


bourgeoa:
I used the /dist/window/solid-file-client.bundle.js. Hope it is the good one ?

Otto-AA:
> I used the /dist/window/solid-file-client.bundle.js. Hope it is the good one ?

Yes, that's the browser version.

Do you know what the res in [this line](https://github.com/jeff-zucker/solid-file-client/pull/65/files#diff-92c0ef5cb9c19d514737574886f31813R23) is doing? It is not initialized and I don't think it is used anywhere.

In line 69, if you change the test behavior, you should also change the description. And @jeff-zucker , when should updateFile return 404? Shouldn't it always create the new file, no matter if the parents exist?

jeff-zucker:
@bourgeoa great work, thanks.  We should not have two versions of the test.  You should call the node version from within the html with a script tag.  That may require some modification of the node version, I'll look.

I would prefer that he folder for the browser version to be called dist/browser. and for the node version dist/node.  The examples likewise should be browser-example.html and node-example.js.

The res at the top is not used in the code currently, but it allows me to stick in a console.log(res) anywhere I want without needing to worry about whether or not I've previously declared it with let.  We can remove it when the tests are finalized.  In general,
my coding style is to leave various stubs and tests in until I am done and ready to refactor and cleanup.  So running lint on my code before then is not likely to be very useful.

You are correct that if the parent folder is not found, updateFile() should create it.  That is not what that test is testing.  It is looking for a file on a different server that doesn't exist.  It says "file not found" not "parent folder not found". Probably an unnecessary test that should be replaced by one that tests what you are talking about.

PR #67, opened 2019-06-30
crud-browser uses crud-node script

bourgeoa:
I made changes to crud-node (testing (typeof window) as to be able to add it as a script in crud-browser.
There is now only one set of crud tests.

The crud-browser tests passes the same in  v4.3 (bourgeoa.ga) and v5.1.6 (solid.community)

in node the crud-node tests shows : 
- v4.3 : 2 errors in `app://ls`
READ  FAIL query : expected : true got : false
DELETE FAIL 404 delete folder, not found : expected : 404 got : 200

- v5.1.6
-- same 2 errors in `app://ls`
-- 10 errors in `https://` : mostly error 500 in create file


jeff-zucker:
Great work, thanks.

There are problems with how app:// handles containers, I am tracking it
down but it may take me a while.

The problems with createFile are probably caused by this issue I raised:
https://github.com/solid/node-solid-server/issues/1245.  The new NSS
ignores the extension of resources and defaults to text/plain which means
you can PUT foo.ttl and it will get served as text/plain, i.e. will not
respond to queries addressed to its Turtle.  We should make "text/turtle"
the default content-type for the create methods but allow users to
over-ride.  This whole issue will really mess-up Solid IDE so once we have
a response from the NSS devs as to whether this is the new intended
behavior, we'll have to check it.

bourgeoa:
Does your explanation covers the fact that the https:// errors do not exist in crud-browser tests ?

jeff-zucker:
I'm not sure what you mean.  Are you saying that the same tests work in the
browser but not in node?  If so, that's very odd and I'd appreciate help in
tracking it down.

jeff-zucker:
How did you authorize the origin or the browser test on 5.x?

jeff-zucker:
Making createItem default to content-type "text/turtle" fixes the createFile issues.  There are still some 500s, I'm loooking.

PR #68, opened 2019-06-30
some errors and improvements in crud-node

bourgeoa:
I made some changes to `crud-node`. Here is what I got in 

> Preparing test
>   deleting test-folder/default/badurl.ttl
>   deleting test-folder/default/cbc2de30-9b77-11e9-8e5f-b9cca92d8942-test.ttl.ttl
>   deleting test-folder/default/d07d0ae0-9b77-11e9-8e5f-b9cca92d8942-test.ttl.ttl
>   deleting test-folder/default/test.ttl.ttl
>   deleting test-folder/default/
>   deleting test-folder/default/noSuchThing/
>   deleting test-folder/default/deep-folder/deep-file.ttl.ttl
>   deleting test-folder/default/deep-folder/
>   deleting test-folder/default/
>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,7 +59,7 @@
           "integrity": "sha512-pYAIzeRo8J6KPEaJ0VWOh5Pzkbw/RetuzehGM7QRRX5he4fPHx2rdKMB256ehJCkX+XRQm16eZLqLNS8RSZXZw==",
           "dev": true,
           "requires": {
-            "ms": "^2.1.1"
+            "ms": "2.1.2"
           }
         },
         "ms": {
@@ -826,7 +826,7 @@
           "integrity": "sha512-pYAIzeRo8J6KPEaJ0VWOh5Pzkbw/RetuzehGM7QRRX5he4fPHx2rdKMB256ehJCkX+XRQm16eZLqLNS8RSZXZw==",
           "dev": true,
           "requires": {
-            "ms": "^2.1.1"
+            "ms": "2.1.2"
           }
         },
         "ms": {
@@ -3159,7 +3159,7 @@
           "integrity": "sha1-9tXREJ4Z1j7fQo4L1X4Sd3YVM0s=",
           "dev": true,
           "requires": {
-            "find-up": "^2.1.0"
+            "find-up": "2.1.0"
           }
         }
       }
@@ -3207,8 +3207,8 @@
           "integrity": "sha1-N53Ocw9hZvds76TmcHoVmwLFpvo=",
           "dev": true,
           "requires": {
-            "esutils": "^2.0.2",
-            "isarray": "^1.0.0"
+            "esutils": "2.0.2",
+            "isarray": "1.0.0"
           }
         },
         "find-up": {
@@ -3226,10 +3226,10 @@
           "integrity": "sha1-eUfkIUmvgNaWy/eXvKq8/h/inKg=",
           "dev": true,
           "requires": {
-            "graceful-fs": "^4.1.2",
-            "parse-json": "^2.2.0",
-            "pify": "^2.0.0",
-            "strip-bom": "^3.0.0"
+            "graceful-fs": "4.1.15",
+            "parse-json": "2.2.0",
+            "pify": "2.3.0",
+            "strip-bom": "3.0.0"
           }
         },
         "locate-path": {
@@ -3287,7 +3287,7 @@
           "integrity": "sha1-8BLMuEFbcJb8LaoQVMPXI4lZTHM=",
           "dev": true,
           "requires": {
-            "pify": "^2.0.0"
+            "pify": "2.3.0"
           }
         },
         "pify": {
@@ -3302,9 +3302,9 @@
           "integrity": "sha1-jvHAYjxqbbDcZxPEv6xGMysjaPg=",
           "dev": true,
           "requires": {
-            "load-json-file": "^2.0.0",
-            "normalize-package-data": "^2.3.2",
-            "path-type": "^2.0.0"
+            "load-json-file": "2.0.0",
+            "normalize-package-data": "2.5.0",
+            "path-type": "2.0.0"
           }
         },
         "read-pkg-up": {
@@ -3313,8 +3313,8 @@
           "integrity": "sha1-a3KoBImE4MQeeVEP1en6mbO1Sb4=",
           "dev": true,
           "requires": {
-            "find-up": "^2.0.0",
-            "read-pkg": "^2.0.0"
+            "find-up": "2.1.0",
+            "read-pkg": "2.0.0"
           }
         }
       }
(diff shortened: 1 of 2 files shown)


bourgeoa:
I seems that the file extension ( .ttl) is added even when it allready exists.

jeff-zucker:
The problem is either in NSS 5.x POST or in SolidApi.createFile() because
if I add a createFile method to SolidFileClient which uses PUT instead of
SolidApi.createFile(), all tests pass.

jeff-zucker:
@Otto-AA, looks like NSS 5.x has changed how POST works.  I'm trying to pin down the specifics, but 13 tests in crud-node fail using createFIle with POST and they all pass using a straight PUT instead.  I am not so sure we should be doing all creations with POST anyway.

jeff-zucker:
Part of the problem, that I had solved in the old solid-file-client is that you can NOT send an undefined content-type with a POST, so you need to check if a content-type is passed in and if not, don't set the header.  That is what is causing the 500 errors.

Otto-AA:
I think if we make http requests we have to supply a Content-Type header, else it is invalid ([here](https://github.com/solid/node-solid-server/issues/1165#issuecomment-486229117) it is stated for PUT, I guess it is the same for POST). So we cannot send an undefined content-type in PUT either.

And I'm not really sure on how standard PUT requests are, therefore I'd like to stick with POST for createFile and co. If you have other information on this, I'd be glad to hear it, but the api-spec currently says following:

> An alternative (though not standard) way of creating new resources is to use HTTP PUT.

Which in my interpretation means we shouldn't rely on it to work for all solid servers.
https://github.com/solid/solid-spec/blob/master/api-rest.md#http-put-to-create

Otto-AA:
And just to also have it linked here: https://github.com/solid/node-solid-server/issues/1246

It should follow this design issue: https://www.w3.org/DesignIssues/HTTPFilenameMapping.html
Which means that if the file extension and the content type match, it should not be modified:
image

bourgeoa:
Create file with PUT is declared as non standard.
Don't you think it needs spec clarification.
- non standard as not usual
- non standard because not spec compliant
Personally I understood it as spec compliant but non usually used because not clearly defined for container.

bourgeoa:
I remember Tim asking for file extension as a non spec but compatibility issue with usual filesystem. Could it be the source of that error ?

Otto-AA:
> I remember Tim asking for file extension as a non spec but compatibility issue with usual filesystem. Could it be the source of that error ?

Afaik, the problem is that NSS isn't HttpFilenameMapping Design issue I've mentioned before, despite wanting to implement it. If they fix this, `deep-file.ttl.ttl` namings shouldn't happen in NSS.

But I don't think that there's a guarantee that all solid servers do the file naming this way. I don't know of any spec yet which would allow us to get the name of the newly created file (though there may already be one). So if we want our tests to pass on an arbitrary solid server, I don't think we can get the newly created file name without fetching the parent container.

>  Don't you think it needs spec clarification.

Good point, your interpretation also makes sense to me. I've asked for clarification here: https://github.com/solid/solid-spec/issues/198

bourgeoa:
Point 1 I agree with you that it's not a spec and no test should expect that. But it could probably indicates a bug in NSS 5.1.6

Point 2 on PUT seeing Michael de Jong answer I think we now have a clear understanding, it can be used to create a file.

bourgeoa:
Point 1 I seem to remember that @jeffz wrote that with POST the name of the created ressource was returned into header.location and that it is part of the spec.

jeff-zucker:
> I don't know of any spec yet which would allow us to get the name of the
newly created file We certainly  can depend on getting the location of
things created with POST in the location header without any querying of the
container.  It is in the LDP spec https://www.w3.org/TR/ldp/5.2.3.1 LDP
clients should create member resources by submitting a representation as
the entity body of the HTTP POST to a known LDPC. If the resource was
created successfully, LDP servers
<https://www.w3.org/TR/ldp/#dfn-ldp-server> must respond with status code
201 (Created) and the Location header set to the new resource’s URL.
Clients shall not expect any representation in the response entity body on
a 201 (Created) response.
That's an unambiguous MUST on returning the location in the header.  I use
that feature in the tests on the issue.

The tacking on of .ttl to a file that already has an extension is a problem
but is NOT the one I posted the POST issue about.  Ironically, it is the
only one of the four POSTs I showed which is the same as in NSS 4.x.  The
main problem with the new way of doing it is that is is now impossible to
create an extensionless Turtle file.  I'm not sure how it handles non-RDF
resources.

> if we want our tests to pass on an arbitrary solid server

That seems like an absurd goal to me.  My goal is to not violate
specifications where the specifications are solidified and to give the user
the experience they expect on NSS.  If and when there are other servers, we
can write for them.  But the idea of trying to write for all possible
interpretations of a spec which isn't stable is not a path I want to take.
The spec does say the server can name things however it wants to, so we
have **no** idea what to expect in terms of file-naming practices of future
servers.  Trying to write for all possibilities in not possible.

As far as content-types, I believe we should always default to text/turtle
and strip off the .ttl extension if it is supplied.  If there is ever a
server which doesn't add on the extension we can modify.

One of the reasons I favor PUT (though I am not suggesting we use it
exclusively) is that its naming behavior is much simpler.

Otto-AA:
Ok, thanks for the link to the LDP server spec, I haven't looked there yet. So yes, we can definitely use the Location header to find out the name of the created ressource.

> The tacking on of .ttl to a file that already has an extension is a problem but is NOT the one I posted the POST issue about.

Sorry, I've misread that.
I think that the new v5.x behaviour makes sense, because a content-type is required for POST and PUT afaik, but it should return 400 then. Lets see if this was intended or is a side effect of something.

> The main problem with the new way of doing it is that is is now impossible to create an extensionless Turtle file.

Can't we do that with PUT? But I don't really see what the use case of this would be right now.

>  As far as content-types, I believe we should always default to text/turtle and strip off the .ttl extension if it is supplied. If there is ever a server which doesn't add on the extension we can modify.

If we do that, shouldn't we also do that for other extensions too? e.g. .txt, .html, .png, ...?

> One of the reasons I favor PUT (though I am not suggesting we use it exclusively) is that its naming behavior is much simpler.

Yes, but from what I've heard the PUT behaviour is not that well defined (e.g. whether or not it creates the intermediate containers). If I don't have the feeling that the spec is precise enough, I don't really want to use it as it could break with some server implementations who did different things while following the spec.

jeff-zucker:
As far as content-types, I believe we should always default to text/turtle
and strip off the .ttl extension if it is supplied. If there is ever a
server which doesn't add on the extension we can modify.

> If we do that, shouldn't we also do that for other extensions too? e.g.
.txt, .html, .png, ...?

No, I don't believe so.  We don't want to be in the game of guessing
content-types.  But as you have pointed out, it is not much of a jump to
say "if this has a .ttl extension and a Turtle content type, we don't tack
on .ttl".  I'm also not sure how the other kinds of resources are handled,
I havent' seen foo.html.html or bar.png.png.

> from what I've heard the PUT behaviour is not that well defined (e.g.
whether or not it creates the intermediate containers).

The Solid REST spec certainly supports the creation of intermediate
containers, it's only the LDP spec that doesn't have it.  If we support it
we are not violating any spec, the question is only does the spec demand
this behavior (no).  There is nothing that says a client can't create
intermediate containers regardless of what the server does.

But let me say - I am not against sticking with POST as the main creation
tool - for one thing, it makes it easier to track what has been created
with the location header.

PR #69, opened 2019-07-10
Update copy, move and test cases

Otto-AA:
Tested with app:ls copyFolder, copyFile, move and rename now work. I sadly don't have the time to test everything in the browser too. I expect it to work for regular text files, but I could imagine that there are troubles with binary files like images (it should work in the browser because of blob, but I don't know if/how solid-rest-file supports binary files).


jeff-zucker:
I get 5 failed test suites and 68 failed tests.

jeff-zucker:
That's after  pulling your latest PR.

jeff-zucker:
Nevermind.  I will never understand how npm install works. It kept putting an old solid-rest in place.  When I manually replaced it, the tests all work.  Congratulations and thanks!

PR #71, opened 2019-07-11
Add automated documentation

Otto-AA:
It's not the best of the best and I'm not even sure if it is better than directly reading the source code. But I guess it is better than nothing :)

You don't need to do any special documentation for the low level methods if they aren't needed by your documentation. I think this doc is ok as a reference to get more detailed explanation on what a method does. It would be cool if you make a section somewhere with "If the methods above don't fit your needs, you can also use fetch, get, head, ... directly. To see a list of all available methods you can either go to the api.md doc or take a look at the source code yourself". If you have time, you can make it more detailed, but I don't think it is really necessary for now.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,0 +1,589 @@
+## Classes
+
+<dl>
+<dt><a href="#SolidAPI">SolidAPI</a></dt>
+<dd></dd>
+<dt><a href="#SolidFileClient">SolidFileClient</a> ⇐ <code>SolidApi</code></dt>
+<dd><p>Class for working with the solid API</p>
+</dd>
+</dl>
+
+## Constants
+
+<dl>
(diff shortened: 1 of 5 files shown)


Otto-AA:
And there are some parts where I don't know what type is expected. For instance, the login method. I've added TBD comments there. It would be nice if you could take a look at them and comment here what types it should be (or add it yourself). I can quickly add them at the end of July.

jeff-zucker:
Thanks for this.  I definitely plan on an intro section pointing advanced
users to the advanced methods and options.  You can review how I say it at
the end of the month.

jeff-zucker:
>  And there are some parts where I don't know what type is expected. For instance, the login method.

Lets see, if in the browser it expects an IDP string.  If not in the browser it expects to be empty (in which case it looks for an ENV variable and if not found a config file) or to be a hash containing idp, username, password or to be a string (in which case it is ignored and treated the same as if empty, this is done so browser/node scripts can use an IDP string and work in both).

I'm not sure how to express that in JSDoc :-)

PR #73, opened 2019-08-09
Update README.md

scenaristeur:
two little errors in examples


jeff-zucker:
Thanks much! I'll put these in the new v1.0.0 version I'm writing now.  It should be out in September and will have a few changes that aren't backward compatible but also lots of improvements.

jeff-zucker:
No longer relevant as we no longer support session/login methods.

PR #76, opened 2019-10-13
corrections to readFolder : update solidApi._processStatements and rdf-query.loadFromUrl

bourgeoa:
update to : 
- `_processStatements` to return `type` with either `contentType` or `folder`
- add `{ headers: { "Accept": "text/turtle" }}` to `_fetch` in `rdf-query.loadFromUrl` (needed at podroot for fetch to return index.ttl else fetch returns index.html)

With this updates I made solid-ide working with solid-file-client 1.0.0 with minimal changes (
. load folder.content
. checksession returning webIId, it used to be session.webId)


jeff-zucker:
Thanks very much, Alain.

PR #78, opened 2019-11-11
update for withLinks: true

bourgeoa:
With these changes delete/copy recursively are fully fonctional.
tested with solid-ide-v1 : https://www.bourgeoa.ga/solid-ide-v1

I did not make logical changes, just corrected some bugs.
Made deleteFolderRecursively `withLinks: true` by default.
Add to take out the itemRecord from the linkItems array (was not functional for folder) and enrich _getLinkObject
Tested with https://www.bourgeoa.ga/solid-ide-v1 which is still a work in progress.

There are quite a lot of requests which slow things.

### DIFF
--- a/src/SolidApi.js
+++ b/src/SolidApi.js
@@ -281,9 +281,9 @@ class SolidAPI {
    * @param {string} url
    * @returns {Promise<FolderData>}
    */
-  async readFolder (url) {
+  async readFolder (url, options) {
     try {
-      const res = await this.processFolder(url)
+      const res = await this.processFolder(url, options)
       return res
       /*
       const res = await this.processFolder(url)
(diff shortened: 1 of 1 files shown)


Otto-AA:
Do you know where the discussion is we wanted to wait for? I think there was one about the use of .meta files. And one about automatically deleting .acl files if the parent folder is deleted (hence, deleteFolderRecursively wouldn't need to do this).

Also we should think about what happens when we copy a .acl/.meta file. Because they probably contain data about the path too and probably won't work if they are in a different location...? I haven't thought much about that, but we should consider this.

bourgeoa:
Apparently has a discrepancy with PR #79 and PR #80

Issue #52, opened 2019-06-12
v1.0.0 Testing


Otto-AA:
An issue to discuss testing and keep track of testing for the v1.0.0 release

TODOs:
- [ ] Finish tests for SolidApi.js
- [ ] Finish tests for SolidFileClient.js
- [ ] Finish tests for folderUtils.js
- [x] Finish tests for apiUtils.js
- [x] Make account methods (login, ...) work via mocks
- [ ] Add travis CI before releasing
- [x] Transition to using solid-rest
- [x] transition to being able to test in all three environments - app, file, https
- [x] make sure that solid-rest-file behaves the same as NSS
- [ ] improve solid-rest-file to handle all tests

    - [ ]  sr : check location header
    - [ ] sr : create OPTIONS method
    - [ ] sr : send full Response object
    - [ ] sr : remove PATCH from allow header

- [ ] Implement tests which are unsupported by solid-rest at the moment
- [ ] Test everything in a browser
- [ ] Add binary file tests

If you think that something is done, just add a comment. If something is missing, just add it.

Otto-AA:
> On mock, I may have a different solution (in the long run) - if solid-rest and its descendants respond to login requests with mock data, we can test against logging in to file:// or app:// and that way, only the base URL needs to be changed to run the test as real against a pod or mock against solid-rest mini-pod.

I also intended to use solid-rest for testing as soon as it is available, that was one of the reasons I advocated for splitting up solid-rest-browser. But I don't think that it would solve the mock issue for account methods. Afaik it doesn't provide popupLogin methods and similar, so we would still need to mock them...?

jeff-zucker:
Well the tests are using solid-rest-file which is just the un-broken-apart version of solid-rest + solid-file-storage.  

Right, there is no popupLogin method currently but I could easily add it to solid-rest.

bourgeoa:
Completely newbie to test.  Can you tell me what I shall read/try to understand .(..jest+...Jekyl)

jeff-zucker:
[...]

Issue #49, opened 2019-06-09
v1.0.0 : Behavior of write methods


jeff-zucker:
Eventually I will hook in my .acl/.meta work, I mention it here, along with some other ideas.  Here's what I'd like to see.  (I'm reversing myself on the fail if parent not found issue - post should do that, but createFolder doesn't have to)

 * default behavior of createFolder, copyFolder

    * if parent folder(s) don't exist : create them
    * if folder not found : create it
    * default : if folder found : skip with no error
    * if duplicatesOk flag is set : if found - create, server will create a duplicate
    * defaiut : copy associated .acl and .meta files
    * if noLinks tag is set : do not copy associated .acl and .meta files

  * default behavior of createFile, copyFile

    * if parent folder(s) don't exist, create them
    * if file not found, copy/create it
    * default : if file found delete the item, then replace it
    * if duplicatesOk flag set : if file found : create it, server will create duplicate
    * default : copy associated .acl and .meta files
    * if noLinks tag is set : do not copy associated .acl and .meta files

* default behavior of all delete methods 

    * default :  delete associated .acl and .meta files
    * if noLinks tag is set : do not delete associated .acl and .meta files

bourgeoa:
I have some comments on duplicateOk and .acl. 
Either you don't copy .acl or you edit the content link the the ressource. Else you are locked out.

Basically the content link is a relative circular reference. Isn't i also the way for update/rename of a ressource which I feel should be part of solid-file-client.

bourgeoa:
And also delete methods should not allow delete what I will call pod system ressources as /public /private /profile ....

jeff-zucker:
I agree we should not allow deletion of system resources (profile, typeIndex, etc.).

I'm afraid I do not understand what you mean by content link or what you are proposing in relation to duplicateOk and .acl.

bourgeoa:
What I called the content link is the relative name for example `test.json` of the ressource that is the object in the turtle .acl

```:ReadWrite
    a n0:Authorization;
    n0:accessTo <test2.json>;
    n0:agentClass n1:Agent;
    n0:mode n0:Read, n0:Write.
```

jeff-zucker:
[...]

PR #79, opened 2019-11-12
Implement putFile

Otto-AA:
I've added a putFile method and made it the default for copyFile as discussed here: https://github.com/jeff-zucker/solid-file-client/issues/52#issuecomment-552754225

### DIFF
--- a/examples/copy/index.html
+++ b/examples/copy/index.html
@@ -1,0 +1,210 @@
+<!doctype html>
+<html lang="en">
+
+<head>
+    <!-- Required meta tags -->
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+
+    <!-- Bootstrap CSS -->
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css"
+        integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+
+    <title>Solid Copy</title>
+</head>
+
+<body>
+
+    <div class="container">
+        <div class="py-5 text-center">
+            <h2>Solid Copy</h2>
+            <p class="lead">Copy folders or files within your solid pod or from a public source to your pod.</p>
+        </div>
+
+        <div class="row">
+            <div class="col-md-4 order-md-2 mb-4">
+                <h4 class="d-flex justify-content-between align-items-center mb-3">
+                    <span class="text-muted">Login to your Pod</span>
+                </h4>
+                <div class="logged-out">
+                    <button id="login" class="btn btn-primary">Login</button>
+                </div>
+                <div class="logged-in">
+                    <p class="text-muted">Logged in as <span class="webid"></span></p>
+                    <button id="logout" class="btn btn-secondary">Logout</button>
+                </div>
+            </div>
+            <div class="col-md-8 order-md-1">
+                <h4 class="mb-3">Settings</h4>
+                <form id="copy-form" class="needs-validation">
+                    <div class="mb-3">
+                        <label for="src">Folder/File from your pod or public pod</label>
+                        <div class="input-group">
+                            <div class="input-group-prepend">
+                                <span class="input-group-text">From</span>
+                            </div>
+                            <input type="text" class="form-control" id="src"
+                                placeholder="https://solid.example.org/folder/" required>
+                            <div class="invalid-feedback" style="width: 100%;">
+                                Please specify which item you want to copy.
+                            </div>
+                        </div>
+                    </div>
+                    <div class="mb-3">
+                        <label for="dest">Destination at your pod</label>
+                        <div class="input-group">
+                            <div class="input-group-prepend">
+                                <span class="input-group-text">To</span>
+                            </div>
+                            <input type="text" class="form-control" id="dest"
+                                placeholder="https://solid.example.org/other/" required>
+                            <div class="invalid-feedback" style="width: 100%;">
+                                Please specify where it should be copied to.
+                            </div>
+                        </div>
+                    </div>
+
+                    <hr class="mb-4">
+                    <div class="custom-control custom-checkbox">
+                        <input type="checkbox" class="custom-control-input" id="overwrite-files" checked>
+                        <label class="custom-control-label" for="overwrite-files">Overwrite Files</label>
+                    </div>
+                    <div class="custom-control custom-checkbox">
+                        <input type="checkbox" class="custom-control-input" id="overwrite-folders">
+                        <label class="custom-control-label" for="overwrite-folders">Overwrite Folders</label>
+                    </div>
+                    <hr class="mb-4">
+                    <div class="not-copying">
+                        <button class="btn btn-primary btn-lg btn-block" type="submit">Start Copying</button>
+                    </div>
+                    <div class="copying">
+                        <button class="btn btn-primary btn-lg btn-block disabled" type="submit">
+                            Copying...
+                            <span class="spinner-border spinner-border-sm" role="status" aria-hidden="true"></span>
+                        </button>
+                    </div>
+                    <ul class="logs my-2 list-group"></ul>
+                </form>
+            </div>
+        </div>
+
+        <footer class="my-5 pt-5 text-muted text-center text-small">
+            <!-- TODO -->
+            <p class="mb-1">Implemented using solid-file-client</p>
+            <ul class="list-inline">
+                <li class="list-inline-item"><a href="#">Source Code</a></li>
+                <li class="list-inline-item"><a href="#">Docs</a></li>
+                <li class="list-inline-item"><a href="#">Bugs</a></li>
+            </ul>
+        </footer>
+    </div>
+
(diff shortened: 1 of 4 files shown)


Otto-AA:
I didn't test it manually, but the automated tests for putFile against solid.community and app://ls/ worked. The automated tests for copyFile failed due to https://github.com/jeff-zucker/solid-rest/issues/11, but it should work fine.

bourgeoa:
Remember that PUT will not work with .json files  with the actual NSS 5.2.1, expected with 5.2.2 PR approved

Otto-AA:
> Remember that PUT will not work with .json files with the actual NSS 5.2.1, expected with 5.2.2 PR approved

Thanks for the reminder, it is good to know that. But from my point of view it shouldn't keep us away from publishing it like this. It is a server implementation error and NSS is going to be unmaintained in favor of the inrupt pod-server afaik.

bourgeoa:
I suppose you are right but not for the next 6 months at least.

Otto-AA:
Ok, I didn't follow the discussions latealy. My last status is, that on the [18th Sept.](https://gitter.im/solid/node-solid-server?at=5d821c4728c1df0ed650edf1) it was announced, that people working at Inrupt will switch from NSS to IPS "[...] at some point in the near future". Has there been more recent updates on this issue? Or did we find enough volunteers for NSS?

bourgeoa:
Yes in way Inrupt IPS has been postponed.

Jaxoncreed is working back on NSS, and it seems that Inrupt pod server with
0.8 spec has been stopped waiting 1.0

Activity is mainly on NSS, Mashlib, rewriting of tests, communica, rdflib

And specifications with 2 milestones : December 2019 for 1.0 and June 2020

So my assumption for 6 months +
After there may or may not be a migration of data/app.

Following all issues they do not seem to be related to CRUD, but to
restitution/access.

My personal opinion is to finish solid-file-client quickly on NSS nearly as
is. NSS 5.2.2 could be our reference with PUT fix or 5.2.1 without.

I shall try to think about copyFolder.
Could your acl app help diagnose problems ? A kind of audit.

By the way I have added an option to display .acl and .meta files with
solid-ide-v1.
That helps a lot for real tests (having also at the same time mashlib to
edit acl, content editor and content type hovering on filename)
Nota : as there is no browser version of npm mime-type, I am using npm mimer

Otto-AA:
With the https://github.com/jeff-zucker/solid-auth-cli/pull/5 PR the automated tests for copyFile, copyFolder and move also work. So it is able to copy files and folders on the solid.community server. (Only tested with .ttl files. Not with .json and binary files)

Otto-AA:
@bourgeoa Thanks for the summary, it is really helpful.

> I shall try to think about copyFolder.
Could your acl app help diagnose problems ? A kind of audit.

acl-utils is not 100% finished, maybe it could help in the future. But I think we should first discuss, what we want it to be like. If we want to modify the file when we "copy" it, or just stupidly make an identical copy.

Otto-AA:
I've added a fix for the type detection in processFolder. Now it appends to the array instead of overwriting it each time

bourgeoa:
I need your help to exploit the errors results from copy/delete folder.
I want to access to status, status test, URL
The errors can come from to different levels : 
- exemple 403 Access forbidden
- 400, 413 PUT error

We may not use PUT because actually .jpeg do not work. I suppose it is for all audio/video.
I haven't yet report the issue.

bourgeoa:
I think I have a strategy for copy folder.
1. Delete destination
2. Destination folder shall be read-write-control for everybody.
3. Directly create it with acl or check it can inherit from parent
4. Use a copyFolder modified to work with 3 steps : 
- copy folders without .acl (folders inherit acl)
- copy files including .meta, .meta.acl and files.acl without folders.acl
- copy folders.acl files. It must be processed in hierarchical order like folder creation

What do you think ?

Otto-AA:
> We may not use PUT because actually .jpeg do not work. I suppose it is for all audio/video.
I haven't yet report the issue.

For me it did work. Here is what I tried with the copy method from the putFile branch:
```javascript
await fileClient.copy('https://otman.inrupt.net/public/Revive%2Bpost.jpeg', 'https://otman.solid.community/public/revive.jpeg')
```
And it worked without an error (you can take a look if you want).

By the way, I've tried to copy the plume app from one pod to another with the new copyFolder and it worked flawlessly for me. This included png, svg, html, css, js and (as NSS is now at 5.2.2) also json.

Otto-AA:
> I need your help to exploit the errors results from copy/delete folder.
I want to access to status, status test, URL

If you have the https://github.com/jeff-zucker/solid-file-client/pull/80 merged:
From the top of my head (ie not tested) something like this should work:
```javascript
try {
  const results = await fileClient.deleteFolderRecursively(folderUrl)
  results.forEach(response => console.log(`${response.status} ${response.statusText} ${response.url}`))
} catch (err) {
  if (!Array.isArray(err)) {
    // Unexpected error due to a mistake in our code (not because the fetch didn't work
    throw err
  } else {
    // One or more of the requests weren't successful
    // With PR 80 this will be an array of all rejected ones. Without only a single one
    err.forEach(response => console.log(`${response.status} ${response.statusText} ${response.url}`))
  }
}
```

Does this help you...?

Otto-AA:
> What do you think ?

I will come back to this later today.

bourgeoa:
1. My .jpeg error was something like ERR_HTTPS/2_ERROR with PUT.
My server proxy uses nginx. I did allow the HTTPS/2 option.

2. I still have 413 content to big with some .json files and PUT. No problem with POST.

Otto-AA:
Have you also tried it on solid.community or inrupt.net? For me it worked there.

Regarding the 413, it seems weird that it happens on PUT, but not on POST. I thought it occurs when we want to send too much data, independent of the method we use. I will take a look at it later too.

Otto-AA:
> I think I have a strategy for copy folder.

>  Delete destination
    Destination folder shall be read-write-control for everybody.
    Directly create it with acl or check it can inherit from parent
    Use a copyFolder modified to work with 3 steps :

>  copy folders without .acl (folders inherit acl)
    copy files including .meta, .meta.acl and files.acl without folders.acl
    copy folders.acl files. It must be processed in hierarchical order like folder creation

Ad "Delete destination":
We currently have an option `overwriteFolders` which gives the user control over this. It defaults to false, so for copyFolder this means that folders will be merged instead of overwritten. The `overwriteFiles` option currently defaults to true, so existing files will be replaced with the newer version.

If we `overwiteFolders` is enabled, deleting the folder recursively is currently the second thing we do. The first thing is to get the top-level contents of the src folder. This prevents us from deleting the destination, if we can't read the src (e.g. when it doesn't exist or we don't have permissions for it). Hence I would keep this the way it is

Ad "Destination folder shall be read-write-control for everybody.":
I don't think that we should require control permissions for all copy calls, only if "withLinks" is set to true.

Ad "Directly create it with acl or check it can inherit from parent":
What do you mean with the first part of this sentence? That we should create the Folder and its acl file, before we continue with the contents? If "withLinks=true" I would agree with that.

Ad "copy folders without .acl (folders inherit acl)":
Why would you wait with the acl file? From my point of view, they should be created as closely together as possible to keep the window where it is "not protected" by an acl file as small as possible. If we wait e.g. 10 seconds, others could access it for 10s, even though they probably shouldn't. Of course they inherit *some* acl, but probably not the one intended for the files it contains.

Ad "copy files including .meta, .meta.acl and files.acl without folders.acl":
As mentioned in the point above, I wouldn't start copying contents of a folder, before the folder has its acl file copied. Also for coping files, the acl file should be copied right afterwards imo.


There are some other points we should keep in mind:
How do we want to handle errors? Some possible ways:
- reset to the previous state: delete all created items and restore the ones they have overwritten
- reset items with failed acl: keep all items, except those for which the acl creation has failed
- keep everything

If I could simply choose, I'd pick the second one. It makes sure, that no unintended access is granted and keeps everything which succeeded. Implementing it cleanly is still hard, as "restoring the previous state" is tricky. We would have to delete those items and restore the ones they have overwritten, which means we would need to store those. I guess we would need to stay with "just delete items for which acl file creation failed" without restoring previous versions.

Otto-AA:
Btw, I think we should move this discussion to the write behavior issue (#49 )

Otto-AA:
I've added an example website which uses the copy function. While doing this, I've noticed two things:

1) It could be useful to have both, all errors and successful responses. For instance if 3/10 requests fail, I'd like to have 3 error responses **and** 7 success responses to show the user what happened. I'm not sure how we could report this in a nice way, so the developer has access to the failed and successful responses. Maybe throwing an object with `{ failed: [errors...], succeeded: [successful...] }`? (Related to #80 )

2) It could also be useful to have an onProgress callback option, which gets called every time a request finished. It's not necessary, but nice to have as a progress indicator. For instance:
```javascript
const results = await fileClient.copy(from, to, {
  onProgress: (res, errorRes) => {
    if (errorRes) {
      console.error(`Error while creating ${errorRes.url}`)
    } else {
      console.log(`Successfully created ${res.url}`)
    }
  }
})
```

What do you think about these ideas?

bourgeoa:
A progress indicator could be useful, but mostly for the app user, to know that things are running - progress bar.
I like the idea

jeff-zucker:
Coming up to speed on this.  I released new versions of solid-rest and solid-auth-cli including cross-fetch.  So, if I'm understanding the above - I should point SolidFileClient.createFile() to SolidApi.putFile() ?

PR #80, opened 2019-11-12
Reject with all failed promises

Otto-AA:
This changes the error handling behavior of `copyFolder`, `deleteFolderContents` and all methods which use those (`move`, `rename`, `deleteFolderRecursively`). Previously they rejected with the first error, now they will wait for all promises to finish and reject with all found errors.

See https://github.com/jeff-zucker/solid-file-client/issues/52#issuecomment-552538851 for more background info

### DIFF
--- a/src/SolidApi.js
+++ b/src/SolidApi.js
@@ -1,11 +1,13 @@
 // import { getParentUrl, getItemName, areFolders, areFiles, LINK } from './utils/apiUtils'
 import debug from 'debug'
 import apiUtils from './utils/apiUtils'
+import promiseUtils from './utils/promiseUtils'
 import folderUtils from './utils/folderUtils'
 import RdfQuery from './utils/rdf-query'
 
 const fetchLog = debug('solid-file-client:fetch')
 const { getParentUrl, getItemName, areFolders, areFiles, LINK } = apiUtils
+const { promiseAllWithErrors } = promiseUtils
 const { _parseLinkHeader, _urlJoin } = folderUtils
 
 /**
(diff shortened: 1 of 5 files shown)


bourgeoa:
Nice I will try to check tomorrow

bourgeoa:
@Otto-AA thanks for your help with errors handling.

I discovered that errors are not recursive (propagated along the hierarchy).
What do you think of : `copyFolder(from, to, errors, options)`
and something like : errors = errors.concat in the code
I tried it, but my coding is too bad

Otto-AA:
I've fixed an error now that the errors array wasn't flattened (e.g. previously `[err1, err2, [err3, err4]]` could happen). Can you check if this fixes the problem you described? If not, can you give me an example where your problem occurs (e.g. a little folder structure, where some requests fail)

bourgeoa:
Thanks it works. I got all errors from 2 different levels.
I shall make more tests.

Very surprisingly `statusText` is void. Here is my code : 

 		try {
		  const results = await fc.copyFolder(from, to, { withLinks: true })
		  results.forEach(response => console.log(`${response.status} ${response.statusText} ${response.url}`))
		} catch (err) {
		  if (!Array.isArray(err)) {
		    // Unexpected error due to a mistake in our code (not because the fetch didn't work
		    throw err
		  } else {
		    // One or more of the requests weren't successful
		    // With PR 80 this will be an array of all rejected ones. Without only a single one
		    err.forEach(response => console.log(`${response.status} ${response.statusText} ${response.url}`))
		  }
		}

	    return results

Otto-AA:
That's strange. Can you log the full response/err? So we see how it looks like, if it is like a response, or something else. And where did you test it? With solid.community, solid-rest or something else?

Otto-AA:
Also, in your code you define `const results` inside the try block and then attempt to access it outside. I guess that `return results` in the end won't work because of that.

PR #81, opened 2019-11-21
withLinks:true update

bourgeoa:
With these changes delete/copy recursively `withLinks: true` are fully fonctional with .meta only for folders.
- I did not make logical changes, just corrected some bugs.
- Made deleteFolderRecursively withLinks: true by default.
- Add to take out the itemRecord from the linkItems array (was not functional for folder) and enrich _getLinkObject
- added options to readFolder

Tested with https://www.bourgeoa.ga/solid-ide-test-v1 which is still a work in progress.

Is there any reason to keep separate processFolder and createFolder ?

### DIFF
--- a/src/SolidApi.js
+++ b/src/SolidApi.js
@@ -317,8 +317,8 @@ class SolidAPI {
    * @param {string} url
    * @returns {Promise<FolderData>}
    */
-  async readFolder (url) {
-    return this.processFolder(url)
+  async readFolder (url, options) {
+    return this.processFolder(url, options)
   }
 
   /**
@@ -356,7 +356,7 @@ class SolidAPI {
     if (typeof from !== 'string' || typeof to !== 'string') {
       throw new Error(`The from and to parameters of copyFile must be strings. Found: ${from} and ${to}`)
     }
-    const { folders, files } = await this.readFolder(from).catch(responseErrToArray)
+    const { folders, files } = await this.readFolder(from, options).catch(responseErrToArray)
     const folderResponse = await this.createFolder(to, options).catch(responseErrToArray)
 
     const promises = [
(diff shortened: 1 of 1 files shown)


Otto-AA:
Is there a specific reason you used the name `withLinks`? A while ago we talked about it and thought about splitting it into `copyMeta` and `copyAcl` (or `withMeta` and `withAcl` to make it usable for copy and delete).

> Is there any reason to keep separate processFolder and createFolder ?

If you mean processFolder and readFolder I guess we can remove one of them. I think it is a leftover from older code.

jeff-zucker:
I am fine with either way of referring to withLinks or copyACL so I'll stay out of that discussion.  Should I hold off merging the PR until you two come to an agreement?

src/SolidApi.js
```diff
@@ -401,8 +401,8 @@ class SolidAPI {
    * @returns {Promise<Response[]>} Resolves with a response for each deletion request
    * @throws {Response[]|Error} if one or more fetch requests failed an array of the responses.
    */
-  async deleteFolderContents (url) {
-    const { folders, files } = await this.readFolder(url).catch(responseErrToArray)
+  async deleteFolderContents (url, options = { withLinks: true }) {  // should delete .acl by default for deletefolderRecursively
```
Otto-AA: Assigning default options this way can lead to unexpected results. For example if one calls `deleteFolderContents(url, {})` the options will be an empty object instead of the default. Also it doesn't scale well if more options are added.

I'd prefer like this:
```javascript
const defaultDeleteOptions = { withLinks: true } // Top of the file
// ...
async deleteFolderContents (url, options) {
  options = { ...defaultDeleteOptions, ...options }
  // ...
}
```
This will copy the default options and then overwrite all specified options.

If you want an example where the current version could be bad:
```javascript
func = (opt = { a: true, b: false }) => console.log(opt)
func() // logs { a: true, b: false }
func({}) // logs {}
func({ a: false }) // logs { a: false }
```

bourgeoa:
1. I shall make the code changes
2. for the name, I did not follow the previous discussion. I did not think about it.
 - I agree we must separate `acl`. And `withAcl` seems a more neutral name to be used for delete and copy
 - always delete/copy `meta`, then we don't need a parameter for `meta`

I shall do all the changes if there are no disagreement

jeff-zucker:
No disagreement from me, thanks much!

PR #82, opened 2019-11-22
update responseErrToArray to HTTP/2

bourgeoa:
response.statusText does not exist anymore in HTTP/2

see https://stackoverflow.com/questions/42401795/with-http-2-only-xmlhttprequest-responses-statustext-property-seems-to-be-us
and HTTP/2 spec

>  Finally, HTTP/2 simply doesn't have this property anymore, @JulianReschke pointed the right link in this other question I hadn't noticed before: greenbytes.de/tech/webdav/rfc7540.html#rfc.section.8.1.2.4.p.2

### DIFF
--- a/src/SolidApi.js
+++ b/src/SolidApi.js
@@ -731,7 +731,7 @@ class SolidAPI {
  * @param {Response|Error} err 
  */
 function responseErrToArray(err) {
-  if (err instanceof Error || !err.status || !err.statusText) {
+  if (err instanceof Error || !err.status) {
     throw err
   } else {
     throw [ err ]


Otto-AA:
It's defined in the [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object, so I think it is independent of the underlying http protocol for our use case (ie when using window.fetch or similar).

I've also tried it with XMLHttpRequest and got a statusText there too:
```javascript
function reqListener () {
  console.log(this.status);
  console.log(this.statusText);
  // Printed: 200    OK
}

var oReq = new XMLHttpRequest();
oReq.addEventListener("load", reqListener);
oReq.open("GET", url);
oReq.send();
```
If it happened to you, that statusText was undefined it would be interesting where it happened. Because I can't reproduce this currently.

jeff-zucker:
So we should keep statusText  as-is for the time being?

Otto-AA:
> So we should keep statusText as-is for the time being?

Depends on how @bourgeoa found out about this. If there is a use case where this causes an error we should likely remove it.

PR #83, opened 2019-11-23
Cleanup

Otto-AA:
I've updated jsdoc documentation and fixed linting errors. I've also added several TBDs

@jeff-zucker I've added several TBD comments, where I think something should be changed, or where I need help on the types for JSDoc. They are just things that I noticed, so feel free to disagree and comment on them if you have a different opinion. You could either start a review in this PR to tell me your thoughts, or just write github comments somewhere. Whatever you prefer :)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -4,15 +4,13 @@
 <dt><a href="#SolidAPI">SolidAPI</a></dt>
 <dd></dd>
 <dt><a href="#SolidFileClient">SolidFileClient</a> ⇐ <code>SolidApi</code></dt>
-<dd><p>Class for working with the solid API</p>
+<dd><p>Class for working with files on Solid Pods</p>
 </dd>
 </dl>
 
 ## Constants
 
 <dl>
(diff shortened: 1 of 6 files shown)


Otto-AA:
By the way, I think we should move several methods used by readFolder to folderUtils.js. The SolidApi class is already huge enough, so splitting this seems reasonable to me.

jeff-zucker:
I have no problem with moving things from SolidApi to utils as you see fit.

jeff-zucker:
Your suggestions generally look good.  I'll merge and then examine them one
by one.

jeff-zucker:
@Otto-AA , you asked why in rdf-query I set this.store rather than just returning store and why I query an empty source.  This allows us to reuse a source file with a new query without re-loading the source.  So if we send query1 to url1, that stores the loaded file in this.store.  Later, when we do query2 without a source, the query uses this.store from the previous run without re-loading it.  I'm not sure we actually use that feature currently but it seems useful enough that I see no reason to remove it.

jeff-zucker:
I've now documented the above in the code.

Issue #48, opened 2019-06-08
v1.0.0 : Error Interface


jeff-zucker:
Sorry to be opening this can of worms @Otto-AA, but if there is ever a time to examine the essentials, it is now.  So, the more I think about it, the more I like the idea of always resolving, never requiring the user to use catch, always returning a response.ok Boolean, status and statusText.  I welcome being talked out of this if it's a bad idea.

Here are my reasons for wanting to switch away from throw/catch.  

1. If the user forgets to use catch, an error causes a big ugly "unhandled rejection" message that doesn't really make clear who forgot to handle it.

2. It's easy (for me at least) to typo when creating a catch, lots of symbols to type

3. If the user is is avoiding "await" for whatever reason, this kind of thing becomes really difficult with catch:
```javascript
  readFile.then( r=>{
    if(r.ok) do_X(r.body)
    else if(r.status==404) do_Y()
    else do_Z()
  })
```
4. anyone familiar with the fetch API already knows this way
@bourgeoa do you think this adds or removes complexity for the learning user?

bourgeoa:
I'm sorry but I don't understand what add or remove complexity.

jeff-zucker:
Currently we use catch like this:
```javascript
  let content = await readFile(url).catch(e=>{show_err(e)})
  if( content ) show(content)
``` 
I am thinking of this instead:
```javascript
  let response = await readFile(url)
  if( response.ok ) show(response.body)
  else show_err(response)
``` 
If we do it the first way, methods will resolve on success and reject on error and the user must catch the error.   In the new way, methods always resolve.  If they find an error, they put it in the response.status and response.statusText and set response.ok to false and resolve that.  Since they never reject, the user doesn't need to catch.  Does this make sense?  If not, please ask again.

jeff-zucker:
The response.body would, in the easy methods, but not he API methods be already transformed so there would as we do now in solid-file-client so there is no need  for the user to do .text()

bourgeoa:
I don't know if my question as any sense but does it as any consequence on promises resolved with Promise.all because it is allways resolved. Apart this what you propose is that for the learning user the fetch response is used in lieu of the usual catch error promise. I like the idea because I think promise added to me a level of complexity.

jeff-zucker:
Promises would work the same.  These are still asynchronous.  The only difference is that the user will need to examine the response object rather than needing to catch the error.

jeff-zucker:
[...]

PR #84, opened 2019-11-23
Implement custom Error class

Otto-AA:
I've started implementing the Custom Error class so it's easier to see how usable they are. (Based on the discussion starting [here](https://github.com/jeff-zucker/solid-file-client/issues/48#issuecomment-557782445))

For the version with FetchError and ComposedFetchError (as in this PR), error handling is like this (from the copy example):
```javascript
// Composed methods (copyFolder, copy, move, deleteFolderRecursively)
try {
    const res = await fileClient.copyFolder(from, to, {
        overwriteFiles,
        overwriteFolders
    })
    res.forEach(response => {
      console.log(res) // Response
      // Could access res.status/...
    })
} catch (err) {
    console.error(err) // ComposedFetchError
    // Using (more or less) human readable error messages
    err.rejectedErrors.forEach(err => {
        console.error(err) // FetchError
        console.error(err.message)
        // Could access err.response.status/...
    })
    // Using the raw responses of rejected requests
    err.rejected.forEach(res => {
      console.error(res) // Response
      // Could access res.status/...
    })
    // Printing the successful requests
    err.successful.forEach(res => {
      console.log(res) // Response
      // Could access res.status/...
    })
}
// Simple methods (createFolder, putFile, ...)
try {
  const res = await fileClient.createFolder(url)
  console.log(res) // Response
  // Could access res.status/...
} catch (err) {
  console.error(err) // FetchError
  console.error(err.message)
  // Could access err.response.status/...
}
```

I'm pretty happy with the functionality it covers (custom and human readable error messages and also access to successful and rejected responses).

Two questions are still up to debate imo:
(1) Should we divide it into two error classes or merge it?
(2) How should we handle errors which don't involve a response?

Ad 1:
Having only one class would make it more consistent and easier to explain and understand. The cost of it would be, that for simple methods we cannot make a shortcut to get the status or other properties and hence have to type a little bit more. For instance we could not write `err.status` instead of `err.rejected[0].status`. I personally think that the advantages outweigh the disadvantages here and we should use only one custom Error class.

Ad 2:
The question here is how we want to handle errors which are not related to API requests. This can be illegal arguments (e.g. [this code](https://github.com/jeff-zucker/solid-file-client/blob/otto-changes/src/SolidApi.js#L418)), checks with itemExists which return true/false and not a response (e.g. [here](https://github.com/jeff-zucker/solid-file-client/blob/otto-changes/src/SolidApi.js#L239)) but also unexpected errors like an undefined variable we didn't check for.
We could (1) integrate all of them into our custom errors, (2) integrate only the expected ones into our custom errors, or (3) always use `new Error(msg)` for them. I haven't thought much about this yet, just wanted to mention it for now and see if you have any thoughts on it.


jeff-zucker:
I am in favor of having the least to explain so the more things that can
all be done the same  way, the better.  So I guess I would say one class
and include the non-API errors in the same way.

Otto-AA:
The user now only gets ComposedFetchErrors. They consist of:
- message (e.g. `404 https://otman.solid.community/public/solid-file-client/examples/copy/not/ - The requested resource could not be found` or `Cannot copy from a folder url to a file url or vice versa`)
- successful (array of responses from successful requests)
- rejected (array of responses from rejected requests)
- rejectedErrors (array of FetchErrors. They consist of `err.message` and `err.response`)

What is slightly odd with this implementation is, that when an error occurs which is not directly related to a requests (e.g. `Cannot copy from a folder url to a file url or vice versa`) it will throw a ComposedFetchError with successful, rejected and rejectedErrors all being empty. So the user cannot assume that rejected has at least length one. I didn't find a better way to do this though.

jeff-zucker:
Is there no way to retrieve the status code itself apart from the message?
That seems wrong - won't users want to check by status e.g. if 404 ...

Otto-AA:
> Is there no way to retrieve the status code itself apart from the message?

Sure there is. `err.response` is a Response object so `err.response.status` is the status.

For example:
```javascript
try {
  const res = await fileClient.copyFolder(from, to, options)
} catch (err) {
  console.error(err.message) // e.g. "403 https://example.org/foo/ - Make sure that the user has access to the resource\n500 https://example.org/bar/ - An internal server error occured"
  err.rejected.forEach(res => console.error(`An error occured...${res.status}`)
}

**Comment by jeff-zucker** [comment]
Ok, sorry, missed that.

**Comment by jeff-zucker** [comment]
I don't know which of your commits caused this, but I now get 8 errors out of 22 tests in all three environments in crud-node.

**Comment by jeff-zucker** [comment]
Probably all the new error reporting, I should be able to fix it.

**Comment by jeff-zucker** [comment]
Perhaps I did not follow the early discussion on this closely enough.  Am I to understand that all errors are now arrays, including simple things like a 404 or on a single file request? What is the rationale for that?

**Comment by Otto-AA** [comment]
> Perhaps I did not follow the early discussion on this closely enough. Am I to understand that all errors are now arrays, including simple things like a 404 or on a single file request? What is the rationale for that?

Not directly an array, but in an error which has an array. So for a rather simple request it would currently be catched like this:
```javascript
try {
  await fileClient.createFile(...params)
} catch (err) {
  if (!err.rejected.length) {
    console.error(err.message) // e.g. "Item already existed"
  else if (err.rejected[0].status === 404) {
    console.error('Parent container did not exist')
  } else {
    // Something else
  }

try {
  await fileClient.readFolder(url)
} catch (err) {
  if (err.rejected[0].status === 404) { // If you expect it to only throw one response
    console.error('xyz')
  }
}
```

The reason for this choice, was that it would be able to include all error types and does not need two Classes. Also many methods we provide can fail in multiple ways. The only ones which are very likely to only fail in one way, are fetch, get, head, and the other basic API requests. Even createItem and readFolder can fail in multiple ways.

Imo it would be rather unsafe to assume that our methods (except for fetch, head, post, etc) fail only simply. Nonetheless it is only a bit more typing for the user if he/she does not care: `err.rejected[0].status` instead of `err.status`.

If you don't agree with this, we can still change this. For example we could add a status property to the error object when there is only one failed response. But I'm not sure this is much better. Probably wort discussing though.

Otto-AA:
Btw, this PR actually was meant for discussion. I didn't expect you to merge it before we came to a consensus.

jeff-zucker:
Thanks for the explanation.  What troubles me is that this means that every single line of error trapping code in every app currently using solid-file-client will have to change.  That's a much bigger backwards-incompatibility than others we've contemplated.  I wonder if this could be in some kind of init setting.  So when they instantiate solid-fc, they can specify single res.status behavior and not have to change their entire scripts.  Future users could skip that and get the more complex behavior.

I see you've deleted the commit for now.  Any idea how I revert the main branch to not include that?  I've already merged and pulled to local.

Otto-AA:
> I wonder if this could be in some kind of init setting. So when they instantiate solid-fc, they can specify single res.status behavior and not have to change their entire scripts. Future users could skip that and get the more complex behavior.

We could use a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) for that. It kind of wraps all methods and then could catch the error if the user changed a setting and modifies it to make it backwards compatible. I can take a look into it if you want. Was the old error behavior to just return the response? And what happened when a different type of error ocured?

jeff-zucker:
How critical to this release is this change in error-handling?  There  are
so many other things I need to address, is there any way it can wait until
a later release?

BTW, since latest commit, I can no longer run npm install - I get "call
stack exceeded"

jeff-zucker:
I re-cloned and re-installed and now the jest tests work, 7 of 7 passed.  Fantastic.  The crud-node tests will presumably work once I figure out how to revert the error PR.

jeff-zucker:
I added the below to crud-node and now it passes as do all the jest suites.  So, in other words, the new error stuff seems to work fine and probably adding a res.status (which would be the first rejected status) would work for backward compatiblity.
```javascript
 if(typeof(got)==='object' && !got.status) {
     got.status = got.rejected[0].status
  }
```

bourgeoa:
See https://github.com/jeff-zucker/solid-file-client/issues/48#issuecomment-557806144

My initial comments were to always have res.status and use a 424 error for multi-errors.
It is also a way to know that there is more than one and that you can check err.rejected for a list

jeff-zucker:
> See [#48 (comment)](https://github.com/jeff-zucker/solid-file-client/issues/48#issuecomment-557806144)
> 
> My initial comments were to always have res.status and use a 424 error for multi-errors.
> It is also a way to know that there is more than one and that you can check err.rejected for a list

That would work for me.  And possibly always have a statusText which with the 424 could be  "There are multiple errors, check res.rejected" or some such

Otto-AA:
Honestly I am not that happy with a 424 error code. I'd prefer to use a setting for the old error style or just write the first thrown error code there.

> My initial comments were to always have res.status and use a 424 error for multi-errors.
It is also a way to know that there is more than one and that you can check err.rejected for a list

If we want to do it that way, I would prefer not to use a valid http status code. Using 424 might lead to confusion for the users (thinking that the server returned 424 while we did, or the other way round). We could use any other 3-digit number.

Also then the question is what other data we directly assign to the error? Should `err.statusText` also work? And `err.url`, `err.headers`, ... . If err contains a `err.status` field it is intuitive that it also contains other response fields.

And for errors like "Invalid arguments" we would need a different statusText. But I guess that is not a game-changer here.

jeff-zucker:
I guess I am in favor of 1) when there is only a single error, return a standard error response 2) when there are multiple errors, return a composed error that has at least a status and maybe also statusText and maybe the others.  I am slightly in favor of the status being the first error in the case of multiple errors.

Otto-AA:
And for 3) when there is a different error (ie Invalid Arguments) return a normal `new Error`?

If yes, this would mean three different types of error handling if we want to do it properly. Which doesn't seem favorable imo.

> when there is only a single error, return a standard error response

Maybe I misunderstood this. What kind of standard error response do you mean?

jeff-zucker:
> And for 3) when there is a different error (ie Invalid Arguments) return a normal `new Error`?
> 
> If yes, this would mean three different types of error handling if we want to do it properly. Which doesn't seem favorable imo.
> 
> > when there is only a single error, return a standard error response
> 
> Maybe I misunderstood this. What kind of standard error response do you mean?

To simplify : I would like to see a standard Response object with at least tres.status and maybe res.ok and res.statusText on all responses and what that is for multiple reponses and for non-https errors,  I don't really care.

Otto-AA:
If you both agree, I can add status, ok, and statusText to the custom Error class. I am not really convinced that it is the way to go, because I could imagine it to confuse the developer instead of providing them a useful shortcut. But as this is for higher level, I will leave the decision to you.

The error interface would look like this:
```
FetchError
-> name = "SFCFetchError"
-> message = 
---> for single response error: "FetchError 403 https://example.org (+ description for status code. e.g. 'Make sure that the user has access to the resource')
---> for multiple response errors: "Fetch Error 123 https://example.org \n 234 https://other.org (+ description for status codes)
---> for other errors: otherError.message

-> status
---> for response errors: status of (random) first response
---> for other errors: -1
-> statusText:
---> for response errors: statusText of (random) first response
---> for other errors: `Error in SolidFileClient: ${otherError.message}`
-> res.ok = false

-> successful = [list of successful responses]
-> rejected = [list of rejected responses]
-> rejectedErrors = [list of rejected FetchErrors (useful for getting the description of the errors)]
-> errors = [list of all errors, regardless if they have a response or are due to illegal arguments]
```

bourgeoa:
I think it is the way to go except for the error status random first error
status that do not give the information that this not the standard error
but a multiple one.

Otto-AA:
I've updated the message above to include your suggestion, @bourgeoa

jeff-zucker:
I am OK with this error handling.  I presume that in the case a single
error is returned the status, statusText, and ok will all refer to that.

Issue #15, opened 2019-01-14
Question: Can I fetch the foaf:name from the profile using this library?


SQL-MisterMagoo:
Hi, thanks for providing this - It is really helpful.

I wanted to get the name of the logged in user to display in a file management app, and I've been trying to use the fetchAndParse method from my browser console to figure out the right way - with no luck.

I can see the returned object from fetchAndParse(<mywebid>) contains the name I want, but querying the graph is difficult as all the examples use rdflib (like your tests), but that doesn't appear to be available through SolidFileClient or SolidAuthClient.

Would I need to add rdflib to my project to be able to do this or is there some method already available to me?

I tried this, but it returned undefined:

```SolidFileClient.fetchAndParse("https://mistermagoo.inrupt.net/profile/card#me").then(g=>g.any("https://mistermagoo.inrupt.net/profile/card#me",'<http://xmlns.com/foaf/0.1/name>'))```

thanks

**Comment by jeff-zucker** [comment]
Yes, in a round-about way, the solid-file-client library gives you access to rdflib.js.  The thing returned by fetchAndParse() is an rdflib.js graph object, often referred to as kb.  So if you used rdflib.js directly you'd say something like var kb = $rdf.graph(), That is the object that gets returned by fetchAndParse() and you can use it the same as if you'd defined it directly. 
 For example, this code (which should run the same in or out of the browser) gets the foaf name from my profile without needing to invoke rdflib.js directly.  BTW, one problem with your code is that you are trying to use g.any() on literal URLs.  You can't do that in rdflib.js, you need to use the g.sym() to turn the literals into symbols as shown below

```javascript
const fc = SolidFileClient;
let url = "https://jeffz.solid.community/profile/card"
fc.fetchAndParse(url).then( kb => {
    var subj = kb.sym(url+"#me")
    var obj = kb.sym("http://xmlns.com/foaf/0.1/name")
    var names = kb.each(subj,obj)
    console.log(names[0].value)
}, err => console.log(err) );
```

SQL-MisterMagoo:
Thank you so much, that is a big help!

SQL-MisterMagoo:
@jeff-zucker Thanks very much - I got it working and started work on a Solid file browser in Blazor
https://twitter.com/mistermag00/status/1085356004349542400

jeff-zucker:
Congrats! You should announce your app on the forum https://forum.solidproject.org/ and perhaps the glitter chat https://gitter.im/solid/app-development

PR #85, opened 2019-11-23
Add SolidFileClient session mock tests

Otto-AA:
I've added several test cases for the auth part of the SolidFileClient. It does pretty much the same as high-level-session.test.js but mocks the solid-auth-client so it can run in automated tests.

@jeff-zucker Would it be okay for you to rename/delete high-level-session.test.js so it doesn't automatically run with `npm test`? When we do this and upgrade to a new solid-rest version which includes https://github.com/jeff-zucker/solid-rest/pull/15, then all tests should pass again. I think setting up Travis CI would make sense now.

By the way, while making the tests I noticed two things: First, trackSession would be nice to have. Secondly, `popupLogin` calls `login(popupUri)` in a node environment, while I guess that login expects some kind of credentials.

### DIFF
--- a/tests/SolidFileClient.session.test.js
+++ b/tests/SolidFileClient.session.test.js
@@ -1,0 +1,103 @@
+import SolidFileClient from '../src'
+
+const session = { webId: 'me' }
+const credentials = { todo: 'make this more realistic' }
+const defaultPopupUri = 'https://solid.community/common/popup.html'
+
+const mocks = {
+    login: jest.fn(),
+    logout: jest.fn(),
+    currentSession: jest.fn(),
+    popupLogin: jest.fn(),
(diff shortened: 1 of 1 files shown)


jeff-zucker:
@otto_A_A I removed the unused tests and pushed everything I could find,
including solid-rest #15 (makeContainers).  I bumped versions of solid-rest
(1.0.9) and solid-auth-cli (1.0.10).

I am not able to run the jest tests. With npm run test, I get :

Determining test suites to run...Error in setup.js: Couldn't setup the
context
/var/www/html/solid/solid-file-client/src/SolidApi.js:1
import debug from 'debug'
       ^^^^^
SyntaxError: Unexpected identifier

PR #87, opened 2019-11-24
Add Travis CI

Otto-AA:
I've switched back to jest 24.8.0 because the new version causes an error. I will take a look at this when I have more time.

Now all tests pass. Travis runs the tests with the `app://ls/` and `file://` prefixes via solid-rest. One of you would have to enable it on https://travis-ci.org/


Otto-AA:
I've updated the dependencies and jest is now also up to date.

Otto-AA:
I've added a "FileApi" which mimics some methods of the SolidApi for files. So currently it uses this for setting up the tests:
- app://ls/ uses SolidApi for the test setup (hence errors for deleteFolderRecursively and co are harder to debug)
- file:// uses FileApi for the test setup (so I'd suggest to debug errors with deleteFolderRecursively with `npm run test:file`)
- https:// uses SolidApi for the test setup (I'd use this for checking if something is solid.community compliant. It makes a lot of requests so I would only do it for a subset of the tests. e.g. add a .only to the describe block you want to test (`describe.only('deleteFolderRecursively', () => ...`) and only run one test file (`npx jest tests/SolidApi.composed.test.js`))

jeff-zucker:
I do not understand the rationale for having two different APIs for the
three tests. Can you explain?

jeff-zucker:
I am getting unresolved conflicts when trying to push this.  @Otto-AA can you resubmit?

Otto-AA:
For instance if we take this test for `deleteFolderRecursively`:
```javascript
test('after deletion itemExists returns false on folder and all contents', async () => {
  await api.deleteFolderRecursively(parentFolder.url)
  await expect(api.itemExists(parentFolder.url)).resolves.toBe(false) // TODO: parentFolder exists in app://ls/ environment
  await Promise.all(parentFolder.contents.map(item => expect(api.itemExists(item.url)).resolves.toBe(false)))
})
```
Here I check that after calling deleteFolderRecursively on a folder, it really is deleted and also the contents of it are gone.. To be able to do that, it is useful to have an automated building of the folder structure, so we can quickly add new tests.

For the automated building the TestFolderGenerator (without this PR) relies on SolidApi to create this folder structure. In particular it also uses SolidApi.deleteFolderRecursively to reset the structure between each test. The problem with this is, that when SolidApi.deleteFolderRecursively somehow fails, none of the tests will be able to run. So we would have to manually find out why it fails. We can't check which tests still pass to narrow down where we have to search for the error (e.g. if readFolder is still working, or if deleteFolderContents is working, etc).

FileApi provides a way to build the folder structure independent of the SolidApi methods. If we introduce a bug in SolidApi.deleteFolderRecursively we would still see what tests pass and should be able to quickly find the cause. But as the name suggests, it only works for the file system, so `app://ls/` and `http://` won't work with it. That's why I think it would be good to have 2 building Apis. I hope I was clear with the explanation :))

jeff-zucker:
Thanks for the detailed explanation, makes sense.

PR #101, opened 2019-12-13
Otto changes

bourgeoa:
This PR added a property `links exists` to `readFolder()`
This property is used in the proposed solid-ide [https://github.com/bourgeoa/solid-ide](url)
offering access to` .acl` attached to file or folder and to be edited in `databrowser`


Otto-AA:
After merging this PR I definitely like to refactor some parts of how readFolder and the link utils work. Currently they rely on being attached to SolidApi so the `this` of them has several methods and I think this is very confusing and not a good practice.

Apart from that I would try to make those methods not require a fetch but getting a response object as parameter. So the current `readFolder` would be based upon a `readFolderFromResponse` method and similar for the link utils. The links refactor could reduce the amount of requests we make, the other just decouple folder utils and SolidApi a bit.

bourgeoa:
Can you do it on my fork.

I remember your comments but did not understand what to do nor how to do it

jeff-zucker:
> Can you do it on my fork.

So, I'm unclear, should I push this PR #101 or wait till you've refactored?

bourgeoa:
The easiest is to push PR #101 in all cases

Otto-AA:
I will make a PR with the changes. It should be finished Monday "evening" or earlier.

jeff-zucker:
Excellent!  I will post the docs for all this later today.  I'd much appreciate it if you could give those a once over to make sure the docs match the code.

PR #106, opened 2019-12-21
Refactor

Otto-AA:
This is not yet completely finished and not tested at all. It's just so you see what changes I'd make and you can comment on them. If you want to merge this, it shouldn't break much, but I guess that the acl and links stuff isn't working yet.

Here are the main changes:

### rdf-query
I've slightly modified the structure so it can cache multiple urls. The cache is an object with urls as keys and N3Stores as values. The only change which isn't 100% compatible with the previous version is, that it requires the url on all queries (previously not necessary when cached).

- query
  - accepts useCache=true option
- queryTurtle can be used if the turtle is already available
- _queryCached loads the store from the cache.

### linksUtils
Now independent of the fetch method, ie the methods now has the response/link header of a request as parameter.

- parseLinkHeader (linkHeader, itemUrl)
  - returns obj of links (e.g. `{ acl: './foo.acl', meta: './foo.meta' }
- getLinksFromResponse (response, [url]) forwards to parseLinkHeader

At least for me the way it works is now much less confusing and simpler than before. I hope it's the same for you.

### folderUtils
Now folderUtils is independent of the fetch method, it now accepts a response as parameter.

- parseFolderResponse (folderResponse, folderUrl)
  - returns the same as processFolder previously
  - doesn't handle links of contents (only for the folder itself)

#### Creating files
Some changes to reflect what we said in the background discussion:

- postItem (previously createItem)
  - accepts createPath=true option
- postFile (calls postItem)
- createFile uses putFile

#### Working with links

- readFolder: In theory the option to include (optionally possible) links has been implemented. Not tested yet so it maybe works, maybe not.
- copy: still need to work this out
- deleteFolder: still need to work this out

#### Merge options
Not yet implemented

### DIFF
--- a/src/SolidApi.js
+++ b/src/SolidApi.js
@@ -232,32 +232,6 @@ class SolidAPI {
       })
   }
 
-  async getLinks (url) {
-    // TODO: Consider making private and/or remove completely. Currently used in folderUtils
-    const res = await this.head(url)
-    const links = getLinksFromResponse(res, url)
-    const linksArr = []
-    if (links.acl && await this.itemExists(links.acl)) {
-      linksArr[0] = {
-        url: links.acl,
-        type: 'text/turtle',
-        itemType: 'AccessControl',
-        name: getItemName(links.acl),
-        parent: getParentUrl(links.acl)
-      }
-    }
-    if (links.meta && await this.itemExists(links.meta)) {
-      linksArr[1] = {
-        url: links.meta,
-        type: 'text/turtle',
-        itemType: 'Metadata',
-        name: getItemName(links.meta),
-        parent: getParentUrl(links.meta)
-      }
-    }
-    return linksArr
-  }
-
   async getItemLinks (url) {
     return this.head(url).then(getLinksFromResponse)
   }
--- a/src/SolidApi.js
+++ b/src/SolidApi.js
@@ -247,7 +247,7 @@ class SolidAPI {
       }
     }
     if (links.meta && await this.itemExists(links.meta)) {
-      linksArr[1] =  {
+      linksArr[1] = {
         url: links.meta,
         type: 'text/turtle',
         itemType: 'Metadata',
@@ -397,7 +397,7 @@ class SolidAPI {
       ...options
     }
 
-    const folderRes = await this.get(url, { headers: { Accept: 'text/turtle' }})
+    const folderRes = await this.get(url, { headers: { Accept: 'text/turtle' } })
     const parsed = parseFolderResponse(folderRes, url)
 
     if (options.links === LINKS.INCLUDE_POSSIBLE || options.links === LINKS.INCLUDE) {
@@ -420,7 +420,7 @@ class SolidAPI {
     const addPossibleLinks = async item => item.links = await this.getItemLinks(item.url)
     await composedFetch([
       ...folderData.files.map(addPossibleLinks),
-      ...folderData.folders.map(addPossibleLinks)      
+      ...folderData.folders.map(addPossibleLinks)
     ])
   }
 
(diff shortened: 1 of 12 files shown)


jeff-zucker:
Thanks so much, great work.

> rdf-query

Looks good.

> parseLinkHeader (linkHeader, itemUrl)
returns obj of links (e.g. `{ acl: './foo.acl', meta: './foo.meta' }

What about ./foo.meta.acl?

Everything else looks good to me.

jeff-zucker:
I merged , pulled, built, and tested - it passes more tests than the last version!

Otto-AA:
> parseLinkHeader (linkHeader, itemUrl)
    returns obj of links (e.g. `{ acl: './foo.acl', meta: './foo.meta' }

Ok, I've tested it and my previous example was not correct. For example a request to a public folder on NSS returns this link header: `<.acl>; rel="acl", <.meta>; rel="describedBy", <http://www.w3.org/ns/ldp#Resource>; rel="type"`. This will be parsed to:
```javascript
{
  acl: 'https://.../public/.acl',
  describedBy: 'https://.../public/.meta',
  type: 'http://www.w3.org/ns/ldp#Resource'
}
```

It will just use every rel as key, and every url as value (while making it absolute).

> What about ./foo.meta.acl?

If we want to include such links, we will need to make an additional request because it is one level deeper. Similar for `foo.acl.acl`. I thought that the first level would be enough. If you think that we should always make 2 requests instead and include .meta.acl we could do that. Or 3 requests and `foo.acl.acl` too.

Regarding the keys of the object: Currently it uses `links.describedBy` for meta files. If you want it to be `links.meta` we could manually change that in SolidApi. I'd personally prefer the first as it reflects the link header better, but I don't have a strong opinion on that.

Also note, that the name `fc.getItemLinks` will change in the future. I'd tend to making a `fc.getPossibleItemLinks` and `fc.getExistingItemLinks` method or passing an option to it (like for readFolder).

jeff-zucker:
>   describedBy: 'https://.../public/.meta',

I would keep it as .meta because there is huge and ongoing discussion about what the link should be named and describedBy with a capital B is not one of the contenders.  Even rdflib does not keep describedBy with a capital B.  Also .acl may become "accessTo". This will all likely need to be changed when the new spec is decided on, so I vote for just staying close to what the files are named rather than trying to guess what he vocabulary will end up being.      OTOH, I don't feel strongly about this, so whatever you decide is okay.

> > What about ./foo.meta.acl?
> 
> If we want to include such links, we will need to make an additional request because it is one level deeper.  

True.

> Similar for `foo.acl.acl`. 

As far as I know, that kind of link is not supported in the spec or won't be in the next spec as the consensus seems to be that it would lead to infinite recursion.  I do not recommend supporting it, then we'd need to look for foo.meta.acl.acl and other absurdities.

> If you think that we should always make 2 requests instead and include .meta.acl we could do that.

Well, we need to do it for deleteFolder(), otherwise the folder deletion will fail because of the remaining foo.meta.acl.  I think we probably should support it.  But it's such a minor possibility, I'm okay with not doing the 2nd request for now.

> Also note, that the name `fc.getItemLinks` will change in the future. I'd tend to making a `fc.getPossibleItemLinks` and `fc.getExistingItemLinks` method or passing an option to it (like for readFolder).

I'd prefer getItemLinks() with a flag rather than two methods, but no strong opinion.

bourgeoa:
I made some tests in solid-ide and taking in consideration that withAcl=true is not implemented, I had the following results for readFolder : 
- the top folder has always 3 possible links acl and describedBy and type links, the other folder in `folders` have no links
- with options.links='includeLinks' or 'includePossibleLinks' error : `TypeError: Cannot read property 'map' of undefined`

When I did the previous withAcl=true links were limited to acl and describedBy (meta)
URI` .meta` was treated as having a dual quality : being a link (the URI is POD defined) and having file property (acl, user defined turtle content)

Otto-AA:
> the top folder has always 3 possible links acl and describedBy and type links

Yes, we need to make a request to the top folder to get its turtle, so we know its links in all cases.

> with options.links='includeLinks' or 'includePossibleLinks' error : TypeError: Cannot read property 'map' of undefined

Fixing that. I forgot to await a promise.

> When I did the previous withAcl=true links were limited to acl and describedBy (meta)
URI

It's an easy change to limit it to those two, if desired. Especially when only including existing links it makes sense (so for example we don't check if `type: 'http://www.w3.org/ns/ldp#Resource'` exists).

> .meta was treated as having a dual quality : being a link (the URI is POD defined) and having file property (acl, user defined turtle content)

Thanks for pointing that out, I missed this change. We can also add acl and meta to the files array. But I'd tend not to do that as it doesn't reflect the atomic model which probably will come.

PR #107, opened 2019-12-22
Update readFolder links

Otto-AA:
readFolder now works with the flags (exclude as default, include and includePossible). I've added an automated test which covers some of the behavior and also tried it out manually in the browser.

Questions that remain open:
- If the top folder has a meta file NSS adds it to the files. So existing meta files currently are in links and files. I'd leave it like that.
- Should we fail if a request to a link or file in the folder fails? In particular to get possible or existing links. Currently I've implemented it to ignore errors when checking if a link exists, errors when getting possible links are thrown I think. I didn't think much about this though.

Usage:
```javascript
const { LINKS } = SolidFileClient
await fc.readFolder(url) // defaults to LINKS.EXCLUDE
await fc.readFolder(url, { links: LINKS.EXCLUDE })
await fc.readFolder(url, { links: LINKS.INCLUDE })
await fc.readFolder(url, { links: LINKS.INCLUDE_POSSIBLE })
await fc.readFolder(url, { links: SolidFileClient.LINKS.EXCLUDE })
// Also possible but not suggested because it's easier to make typos: 
await fc.readFolder(url, { links: 'excludeLinks' })
```

### DIFF
--- a/src/SolidApi.js
+++ b/src/SolidApi.js
@@ -11,12 +11,12 @@ const { FetchError, assertResponseOk, composedFetch, toFetchError } = errorUtils
 const { getLinksFromResponse, parseLinkHeader } = linksUtils
 const { parseFolderResponse } = folderUtils
 
-const MERGE = {
+export const MERGE = {
   REPLACE: 'replace',
   KEEP_SOURCE: 'source',
   KEEP_TARGET: 'target'
 }
-const LINKS = {
+export const LINKS = {
   EXCLUDE: 'exludeLinks',
   INCLUDE: 'includeLinks',
   INCLUDE_POSSIBLE: 'includePossibleLinks'
--- a/src/SolidApi.js
+++ b/src/SolidApi.js
@@ -405,7 +405,7 @@ class SolidAPI {
     const links = await this.head(url).then(getLinksFromResponse)
 
     if (options.links === LINKS.INCLUDE) {
-      this._removeInexistingLinks(links)
+      await this._removeInexistingLinks(links)
     }
 
     return links
--- a/src/SolidApi.js
+++ b/src/SolidApi.js
@@ -51,11 +51,18 @@ const defaultSolidApiOptions = {
 }
 
 /**
- * @typedef Item
+ * @typedef {object} Links
+ * @property {string} [acl]
+ * @property {string} [meta]
+ */
+
+/**
+ * @typedef {object} Item
  * @property {string} url
  * @property {string} name
  * @property {string} parent
  * @property {"Container" | "Resource"} itemType
+ * @property {Links} [links]
  */
 
 /**
@@ -63,6 +70,7 @@ const defaultSolidApiOptions = {
  * @property {string} url
  * @property {string} name
  * @property {string} parent
+ * @property {Links} links
  * @property {"folder"} type
  * @property {Item[]} folders
  * @property {Item[]} files
@@ -234,10 +242,6 @@ class SolidAPI {
       })
   }
 
-  async getItemLinks (url) {
-    return this.head(url).then(getLinksFromResponse)
-  }
-
   /**
    * Create an item at target url.
    * Per default it will create the parent folder if it doesn't exist.
@@ -365,7 +369,6 @@ class SolidAPI {
    * @returns {Promise<FolderData>}
    * @throws {FetchError}
    */
-
   async readFolder (url, options) {
     url = url.endsWith('/') ? url : url + '/'
     options = {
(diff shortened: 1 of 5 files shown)


jeff-zucker:
> * If the top folder has a meta file NSS adds it to the files. So existing meta files currently are in links and files. I'd leave it like that.

That's fine.  Cleanest, would be to have it only in links, but not a big deal.

> * Should we fail if a request to a link or file in the folder fails? In particular to get possible or existing links. Currently I've implemented it to ignore errors when checking if a link exists, errors when getting possible links are thrown I think. I didn't think much about this though.

For readFolder(), we should never fail if a possible link doesn't exist - this is the most common case and does not indicate an error.   We should fail if we can't read a possible location - that should never happen.  We should never fail if a file is listed as being contained in the container but it is not accessible to the user - again this is not an error.  We should fail if a file is listed as being contained but we can't read the file for some other reason.

bourgeoa:
> > * If the top folder has a meta file NSS adds it to the files. So existing meta files currently are in links and files. I'd leave it like that.
> 
> That's fine. Cleanest, would be to have it only in links, but not a big deal.

I did not find the .meta file so I suppose you are both saying the same thing. It is not in array files.
Will see the impact when withAcl=true. and looking for .meta.acl

I have found links property in array folders.
Do you think it is really needed ? These links are one level down the tree. I think we could have the links only on the same container level, that is folder and array files. It will also save performance.

Otto-AA:
@bourgeoa While testing I got following result which includes the .meta file in the files list. I don't know if it is the common case though, as it also says that the type is "application/octet-stream" and I never actively created it...
If you have an example where it behaves differently, could you share the url to it?

```javascript
await fileClient.readFolder('https://otman.solid.community/public/solid-acl-utils/examples/')
{
  "type": "folder",
  "itemType": "Container",
  "name": "examples",
  "parent": "https://otman.solid.community/public/solid-acl-utils/",
  "url": "https://otman.solid.community/public/solid-acl-utils/examples/",
  "links": {
    "acl": "https://otman.solid.community/public/solid-acl-utils/examples/.acl",
    "meta": "https://otman.solid.community/public/solid-acl-utils/examples/.meta"
  },
  "folders": [
    {
      "url": "https://otman.solid.community/public/solid-acl-utils/examples/share/",
      "type": "folder",
      "itemType": "Container",
      "name": "share",
      "parent": "https://otman.solid.community/public/solid-acl-utils/examples/"
    }
  ],
  "files": [
    {
      "url": "https://otman.solid.community/public/solid-acl-utils/examples/.meta",
      "type": "application/octet-stream",
      "itemType": "Resource",
      "name": ".meta",
      "parent": "https://otman.solid.community/public/solid-acl-utils/examples/"
    }
  ]
}
```

Otto-AA:
> I have found links property in array folders.
Do you think it is really needed ? These links are one level down the tree. I think we could have the links only on the same container level, that is folder and array files. It will also save performance.

The default setting is to exclude links. So usually it only makes one request and does not include links for the contents of the folders. Only if `{ links: LINK.INCLUDE}` or `{ links: LINK:INCLUDE_POSSIBLE }` is set it will make additional requests.

Something that I find slightly not intuitive is, that it includes possible links of the top folder as default. I did this because we get this information without additional requests, but I guess that it could confuse users on whether these are possible or really existing links. What are your thoughts on this? Should I remove this for the default, leave it like that, or make additional requests to check if they exist? I'd prefer one of the first solutions.

jeff-zucker:
> Something that I find slightly not intuitive is, that it includes possible links of the top folder as default. I did this because we get this information without additional requests, but I guess that it could confuse users on whether these are possible or really existing links. What are your thoughts on this? Should I remove this for the default, leave it like that, or make additional requests to check if they exist? I'd prefer one of the first solutions.

I would say remove it for the default, but no strong preference.

bourgeoa:
> Something that I find slightly not intuitive is, that it includes possible links of the top folder as default. I did this because we get this information without additional requests, but I guess that it could confuse users on whether these are possible or really existing links.

I find it confusing too and I'd prefer to have no links at all. An other solution would be to default to includePossibleLinks but I think it has a performance cost for all array files

> > I have found links property in array folders.
> > Do you think it is really needed ? These links are one level down the tree. I think we could have the links only on the same container level, that is folder and array files. It will also save performance.
> 
> (I'm not entirely sure if I understood you correctly. If this doesn't answer your question, please ask again :)
> The default setting is to exclude links. So usually it only makes one request and does not include links for the contents of the folders. Only if `{ links: LINK.INCLUDE}` or `{ links: LINK:INCLUDE_POSSIBLE }` is set it will make additional requests.

To make it short (nothing rough) I will always exclude links from array folders, as I do not see any use case.
All uses case of links (existing or possible) in readFolder(foo/) that I think of are on the container level itself. And not in the next down level : 
- foo/ is see uses case for links foo/.acl and foo/.meta
- but for foo/bar/  links foo/bar/.acl or foo/bar/.meta I do not see any. I shall see them in readFolder(foo/bar/)
- foo/test.ttl links are at the container foo/ level foo/test.ttl..acl so I see uses  case

bourgeoa:
> While testing I got following result which includes the .meta file in the files list. I don't know if it is the common case though, as it also says that the type is "application/octet-stream" and I never actively created it...

I think there is an anomaly in your .meta file It should be text/turtle so I suppose it exists as .meta$.bin
In my cases It never appears in the array files list since your refactoring

I went to your pod and the container turtle contains your .meta as a file (It should not - so it is a bad .meta with a bad contentType)

bourgeoa:
@Otto-AA your access to the pod you created on https://bourgeoa.ga is still active so I can check the filesystem when something like that appears. Feel free to use it for your tests

PR #108, opened 2019-12-22
Update acl and merge behavior

Otto-AA:
I've implemented the merge options for createFolder and putFile and the copy methods simply pass on this option. I've removed overwriteFiles and overwriteFolders because it seemed to confusing to have this and the merge options.

Usage:
```javascript
const { MERGE } = SolidFileClient
await fc.copy(from, to) // Defaults to replace
await fc.copy(from, to, { merge: MERGE.REPLACE }) // Delete target before copying
await fc.copy(from, to, { merge: MERGE.KEEP_SOURCE }) // Overwrite files in target, keep target folder
await fc.copy(from, to, { merge: MERGE.KEEP_TARGET }) // Will not overwrite existing files

// I made it work this way because it is easy to implement.
// If you want you can document it, but it's not necessary
await fc.putFile(..., { merge: MERGE.REPLACE }) // Default PUT behavior
await fc.putFile(..., { merge: MERGE.KEEP_SOURCE }) // Default PUT behavior
await fc.putFile(..., { merge: MERGE.KEEP_TARGET }) // Throw if target exist, else PUT
await fc.createFolder(url) // Ensures that the folder exist
await fc.createFolder(url, { merge: MERGE.REPLACE }) // Deletes the folder if it exists, then creates a new one
await fc.copyFile(from, to, { merge: REPLACE }) // default
await fc.copyFile(from, to, { merge: KEEP_TARGET }) // Throw if target exist, else copy
```

withAcl should also work for copying and deleting now, but it's not tested at all (neither with automated tests, nor with manual testing). I won't have time for this today, if you want you can test it out.

And readFolder now doesn't include the possible links of the top folder per default.

> To make it short (nothing rough) I will always exclude links from array folders, as I do not see any use case.
All uses case of links (existing or possible) in readFolder(foo/) that I think of are on the container level itself. And not in the next down level :

>    foo/ is see uses case for links foo/.acl and foo/.meta
    but for foo/bar/ links foo/bar/.acl or foo/bar/.meta I do not see any. I shall see them in readFolder(foo/bar/)
    foo/test.ttl links are at the container foo/ level foo/test.ttl..acl so I see uses case

Ok, now I understand what you mean. What are the use cases you see for this? I currently don't see many use cases for readFolder with links in both cases.

### DIFF
--- a/src/SolidApi.js
+++ b/src/SolidApi.js
@@ -34,7 +34,8 @@ export const LINKS = {
 const defaultWriteOptions = {
   overwriteFiles: true,
   overwriteFolders: false,
-  withAcl: false, // TODO: Set true
+  withAcl: true,
+  merge: MERGE.REPLACE,
   copyMeta: true,
   createPath: true
 }
@@ -234,10 +235,7 @@ class SolidAPI {
     return this.head(url)
       .then(() => true)
       .catch(err => {
-        // Only return false when the server returned 404. Else throw
-        if (err.status !== 404) {
-          throw err
-        }
+        assertResponseStatus(404)(err)
         return false
       })
   }
@@ -296,9 +294,7 @@ class SolidAPI {
       }
       await this.deleteFolderRecursively(url)
     } catch (e) {
-      if (e.status !== 404) {
-        throw e
-      }
+      assertResponseStatus(404)(e)
     }
 
     return this.postItem(url, '', 'text/turtle', LINK.CONTAINER, options)
@@ -397,7 +393,7 @@ class SolidAPI {
    * @param {object} [options] specify if links should be checked for existence or not
    * @returns {Promise<Links>}
    */
-  async getItemLinks (url, options = { links: INCLUDE_POSSIBLE }) {
+  async getItemLinks (url, options = { links: LINKS.INCLUDE_POSSIBLE }) {
     if (options.links === LINKS.EXCLUDE) {
       toFetchError(new Error('Invalid option LINKS.EXCLUDE for getItemLinks'))
     }
@@ -421,7 +417,7 @@ class SolidAPI {
     await composedFetch(
       Object.entries(links)
         .map(([type, url]) => this.itemExists(url)
-          .catch(err => false)
+          .catch(assertResponseStatus(404))
           .then(exists => {
             if (!exists) {
               delete links[type]
@@ -450,14 +446,16 @@ class SolidAPI {
     }
 
     // Copy File
-    const getResponse = await this.get(from).catch(toFetchError)
+    const getResponse = await this.get(from)
     const content = await getResponse.blob()
     const contentType = getResponse.headers.get('content-type')
-    const putResponse = await this.putFile(to, content, contentType, options).catch(toFetchError)
+    const putResponse = await this.putFile(to, content, contentType, options)
 
     // Optionally copy ACL File
     if (options.withAcl) {
+      // TODO: Copy meta. Potentially double writing due to it being listed in files and being a link
       await this.copyAclFileForItem(from, to, options, getResponse, putResponse)
+        .catch(assertResponseStatus(404))
     }
 
     return putResponse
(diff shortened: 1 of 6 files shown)


bourgeoa:
@Otto-AA @jeff-zucker 
beware there is a nasty bug that will lock you out of your folder
SolidApi line 482-483
```
    const toName = getItemName(oldTargetFile)
    const fromName = getItemName(newTargetFile)

```
should be : 
```
    const fromName = getItemName(oldTargetFile)
    const toName = getItemName(newTargetFile)

```

Otto-AA:
Thanks for catching this and sorry if it broke something for you. Will fix it with the next PR.

Otto-AA:
```javascript
content = content.replace(new RegExp('<' + getRootUrl(oldTargetFile) + 'profile/card#me>'), '<./profile/card#me>')
```

@bourgeoa I would remove this line because I think it is not what we want here. The path to the profile is only used to specify the agent and if we move the file, we still want it to point to the same agent.

Otto-AA:
And is there a reason why you separated it into either (exchange absolute path with relative) or (exchange name of relative file path with new name)? I feel like we should do both and not one of these.

bourgeoa:
>  I would remove this line because I think it is not what we want here. The path to the profile is only used to specify the agent and if we move the file, we still want it to point to the same agent.

This is only making non relative to relative for 2 different pod's and I think it is what we want - always relative If .acl could have been. So we make relative the pod's webId if we use different pod's

> And is there a reason why you separated it into either (exchange absolute path with relative) or (exchange name of relative file path with new name)?

If it not relative we make it relative to the newName and there is no need to redo it
When we are already relative we only need to change oldName to newName if they are different

If you want to change we must be sure that the result is correct from one pod to the same pod for copyFile and copyFolder, and the same from one pod to an other one.

To keep something absolute in acl when it could be relative is against the actual working of databrowser in editing acl. I have chosen to make the changes only when needed - minimizing changes, but we could decide to make always relative. I also think @jeff-zucker expressed to prefer to minimize the changes in acl.

jeff-zucker:
> I also think @jeff-zucker expressed to prefer to minimize the changes in acl.

I'm going to stay out of this one, you two are closer to the implementation details and I'll leave it up to you.

Otto-AA:
> To keep something absolute in acl when it could be relative is against the actual working of databrowser in editing acl. I have chosen to make the changes only when needed - minimizing changes, but we could decide to make always relative. I also think @jeff-zucker expressed to prefer to minimize the changes in acl.

I think that we should only change the things which are necessary for it to work. Making something relative so it is easier to read for humans is something that acl-libraries can do, I don't think it is something solid-file-client should take care of.

> This is only making non relative to relative for 2 different pod's and I think it is what we want - always relative If .acl could have been. So we make relative the pod's webId if we use different pod's

When I'm the owner of a file and copy it from "first.example.org/foo/bar.acl" to "second.example.org/foo/bar.acl" I'd also want to be owner afterwards. If we make the profile relative only the owner of the second pod will have access. So I'd tend to make it absolute or at least keep it as it is.

> If it not relative we make it relative to the newName and there is no need to redo it
When we are already relative we only need to change oldName to newName if they are different

An acl file could contain two rules, one with an absolute and one with a relative path. With the current implementation we would fix one, the relative path would still have the old name. If we replace the `else if` with only `if`, then we would fix both. I've also added tests for this and it seems to work fine with checking both.

bourgeoa:
> An acl file could contain two rules, one with an absolute and one with a relative path. With the current implementation we would fix one, the relative path would still have the old name. If we replace the `else if` with only `if`, then we would fix both. I've also added tests for this and it seems to work fine with checking both.

I think you got a point

> When I'm the owner of a file and copy it from "first.example.org/foo/bar.acl" to "second.example.org/foo/bar.acl" I'd also want to be owner afterwards. If we make the profile relative only the owner of the second pod will have access. So I'd tend to make it absolute or at least keep it as it is.
> 
> >

I understand your point, but think it is problematic not to transfer ownership and control to the new pod's webId. In a pod when the owner has not the control it begins to be much more complex.
A solution to that could be to duplicate the rule (when absolute, keep it and duplicate it to a relative one). It could be a parameter ?
In the mean time I have a very strong preference to make it relative as I did.

bourgeoa:
When I have .meta.acl I cannot delete the folder.
I know it is not wildly used be .meta is not also.
I think we cannot leave it like that
There are different solutions to that, I think of 2 : 
- meta treated as a file in array files
- or when deleting a meta file, getItemLink(.meta, 'includePossibleLinks') and delete .meta.acl
- a third one would be to check and add metaAcl in includeLinks when meta exists.
The one and third can be used for copy, move, rename
I did the first but also have thinked of the third one
The second won't copy .meta.acl

Otto-AA:
Ok, I will make it relative to the pod (`/profile/card#me` instead of `./profile/card#me`). I guess that it could be a security risk to not give the owner of the pod also access. I think that we should point it out in the documentation that we do this, as it definitely changes the meaning of the acl file.

Otto-AA:
> When I have .meta.acl I cannot delete the folder.

It works with the version I'm currently working on. I've also added test cases for it and they pass. Creating a PR soon...

jeff-zucker:
You guys are fantastic. Thanks so much for all your work and especially this last minute stuff.  I'm hoping you can be done in time for some well deserved holidays.  If not, I can do whatever's needed after the first of the year.

bourgeoa:
@Otto-AA It seems to work perfectly
- deleting folders with .meta and .meta.acl
- copyFile with acl containing multiple accessTo

I will experiment in as much detail as possible but only tomorrow.
Have you any thought on duplicating the rule for agent webId (absolute for from and relative) ?
 
Thanks for all you have done. I learned a lot and still do not understand all.

Otto-AA:
>  Have you any thought on duplicating the rule for agent webId (absolute for from and relative) ?

You could check if it is copied from one pod to another (by comparing the hosts/origins). If yes, you could exchange `....com/profile/card#me>` with `...com/profile/card#me>, </profile/card#me>` (using "," here so it uses the same subject and predicate as the original one). This should ensure that the pod owner also has access. I think it's ok to do this when there's a note about it in the description.

PR #109, opened 2019-12-22
Add core-js 3 for babel

Otto-AA:
I've installed core-js 3 and set the config in .babelrc. I've also changed `useBuiltIns` to "usage". If I understood it correctly this means, that babel will take care of adding polyfills of methods we used.

I've also moved n3 to dependencies as it is definitely needed for our program to work.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,8 @@
     [
       "@babel/preset-env",
       {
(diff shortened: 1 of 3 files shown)


jeff-zucker:
After remembering to npm install, test results were :

     14 todo, 111 passed, 125 total

PR #112, opened 2019-12-25
Fix file:// tests

Otto-AA:
`npm run test:file` works again, so travis also succeeds now :)

I wanted to finish this because I already spent some time on getting closer to the error. So this is probably the last proactive PR from me for a while.

And as a side note: If some of the tests fail it doesn't necessarily mean that the code is broken. It's more like a hint to check it again. The test cases could also be wrong. But the test cases show which function seems to be wrong and what behavior of it seems not to work. So it should be easy to check if the code has a failure or the tests are written in the wrong way.

### DIFF
--- a/tests/utils/FileApi.js
+++ b/tests/utils/FileApi.js
@@ -1,5 +1,7 @@
 import fs from 'fs'
 
+// Note: Errors here should be thrown in a way that they work for TestFolderGenerator
+//       For instance, TestFolderGenerator._removeFile checks if the response status is 404, so delete has to return 404 if it didn't exist
 class FileApi {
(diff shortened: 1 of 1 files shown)


jeff-zucker:
Great, thanks.  I've made a number of changes in Solid-rest which will make more of the test:file work.   I added the content-type for each contained resource in a container's turtle listing.  I made sure that .acl and .meta were treated as text/turtle.  And most importantly for the tests, I redid how they create the header.link so that it doesn't produce false possibleLinks like foo.acl.meta and foo.acl.acl and foo.meta.acl.acl.  I'll npm the changes soonish.

PR #113, opened 2019-12-27
update links tests for https:

bourgeoa:
@Otto-AA @jeff-zucker 
the SolidApi.links.test.js are now compliant with https:

Tests passes except one : 

The test `copyFolder without links` need to be modified but I don't know-howto :
```
● recursive › copyFolder › copies folder without links

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

      Array [
    -   Any<String>,
    -   false,
    +   "https://bourgeoa.bourgeoa.ga/public/tests/test-folder/SolidApi-links/copy-folder/target/.meta",
    +   true,
```
 
- /target/meta is created by default by createFolder() with length 0 in the copyFolder(withoutLinks param)
- and we must check that it is not created from /source/meta which has length 12

Except the above there is in SolidApi.composed one fail HEAD for folder returns contentType `application/octet-stream` and it was expected as `text/turtle`

### DIFF
--- a/tests/SolidApi.composed.test.js
+++ b/tests/SolidApi.composed.test.js
@@ -1,4 +1,3 @@
-
 import SolidApi, { MERGE } from '../src/SolidApi'
 import apiUtils from '../src/utils/apiUtils'
 import TestFolderGenerator from './utils/TestFolderGenerator'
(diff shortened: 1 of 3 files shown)


jeff-zucker:
Great! Thanks!  It passes all tests in app:// and file://.

PR #115, opened 2019-12-29
Examples

Otto-AA:
I've moved the examples to the docs folder, so they can be viewed as live examples. I've also modified their links to use the npm versions via delivrjs (rather than using relative links to the non-existent dist folder on github pages). It's already been a long day for me, so please check if this makes sense :)

What remains to be done for this PR (not by me, at least not today):
- Updating links in documentation (*.md) to new path and/or live example on github pages
- Possibly automatically replacing relative src with jsdelivr src (so it's easier to try it out with the local version). Not necessary though and only for developing. So that's not for v1.0.0


jeff-zucker:
> I've moved the examples to the docs folder, so they can be viewed as live examples. 

THANKS!

> I've also modified their links to use the npm versions via delivrjs (rather than using relative links to the non-existent dist folder on github pages). It's already been a long day for me, so please check if this makes sense :)

I'll check.

> What remains to be done for this PR (not by me, at least not today):
> 
> * Updating links in documentation (*.md) to new path and/or live example on github pages

I can do that, probably after I merge all the changes to master, that way I can check the links before doing npm publish.

> * Possibly automatically replacing relative src with jsdelivr src (so it's easier to try it out with the local version). Not necessary though and only for developing. So that's not for v1.0.0

Issue #123, opened 2020-01-06
False error when working with index.html in pod root


jeff-zucker:
As reported in https://forum.solidproject.org/t/announce-much-updated-solid-file-client/2546/14, an error is produced when user tries to copy index.html from or to the root of a pod.  The error is "Cannot read property 'endsWith' of undefined".  The file is correctly copied but the error is displayed.  I tried trapping for defined url in both SolidApi and apiUtils but was not able to stop the error from being shown.  AFAIK this only impacts index.html in the pod root, not other files in the pod root or index.html in other folders.

jeff-zucker:
Apparently a pod's root index.html does not return a link header.  We now only error if someone is copying from a source whose index.html has linked files and they have withAcl=true.

PR #124, opened 2020-01-07
add withMeta=false to copyFile()

bourgeoa:
@jeff-zucker This PR should closes issue #123 
Tested on solid.community, inrupt.dev and bourgeoa.ga
files do not have meta at this stage of NSS
It seems that https://podRoot/index.html as no link to meta in headers

Improvement : check that .meta and .acl do exists in headers issue #125


jeff-zucker:
We solved this with Otto's changes to SolidApi instead.

PR #127, opened 2020-02-22
Fix cjs to esm in rdf-query.js

CxRes:
Changed the cjs require statements to esm imports at the top of the file for consistency.

Moved solid-namespace from depDependencies to dependencies in Package.json


CxRes:
Should `debug` also be dependency (and not devDependency)?

jeff-zucker:
Thanks much!

Issue #128, opened 2020-03-02
readFolder() does not return metadata properties in >=1.0


CxRes:
`.readFolder()` is not returning properties such as `mtype`, `modified`, `size` etc. in versions >=1.0. The Readme still shows that these properties are returned.

Can we add a way to get these properties back, at least under a flag?

Otto-AA:
@jeff-zucker I think I've already mentioned it while still working on it, but maybe it got lost somewhere. 

Regarding this code: https://github.com/jeff-zucker/solid-file-client/blob/a501ae987f7d4287cbe6f53cc057ea4e4b79a7b8/src/utils/folderUtils.js#L54-L61

Is the `else if` at line 58 on purpose? I'd expect an `if` there, because currently the `object = object.replace(/.*#/, '')` in 57 will be ignored. That's probably also what prevents other properties from being added, but I'm not sure.

Otto-AA:
@CxRes I agree that it would be useful. But keep in mind that not all properties are required to exist, so if you want to be compatible with different kind of solid pods you should also have a fallback if they don't exist.

CxRes:
@Otto-AA I am sure this is not a server/pod issue. As I say above, this started in >=1.0. Previously, I was getting this information from the same server (inrupt.net) when using solid-file-client. Your comment to Jeff seems to indicate the same as well.

On the other point, if a property is not required to exist (such as because the solid standard does not require it to be made available) this needs to be demarcated in the documentation, so that users can prepare said fallbacks. From what I can gather from the documentation currently there are is an object with 10 properties that are returned. This could be confusing for new users!

Otto-AA:
@CxRes Yes, it is definitely not the pods fault. The properties from NSS are just ignored/dropped by solid-file-client currently.

mtime and co are not required by the standard, hence solid-file-client also can't guarantee this. This should be reflected in our documentation, currently it's misleading (or too much focused on NSS).

Here is the part of the spec where it's mentioned if you are interested: https://github.com/solid/solid-spec/blob/master/api-rest.md#streams

Otto-AA:
And to be clear: In my opinion we should definitely include this optional metadata. If I remember correctly, this was also @jeff-zucker 's intention, but something went wrong in the implementation.

jeff-zucker:
I'll revisit with the intention of including the metadata where available and not breaking where not available.  I'll also revisit the documentation and see if I can make it this part less confusing.

bourgeoa:
[...]

PR #133, opened 2020-03-13
return metadata properties

bourgeoa:
return metadata properties when available (mtype, modified, size)
tests have been updated for their availability in NSS
issue #128


jeff-zucker:
Thanks much.  My eyes are recovering, but it will be a while before I am back to full steam.

PR #137, opened 2020-03-27
return metadata for head folder

bourgeoa:
see issue #128 
Returns metadata for head folder and some lint adjustments.


bourgeoa:
@jeff-zucker I think you could npm, this PR should close the issue #128

Issue #130, opened 2020-03-13
404 Error on local file system access


CxRes:
I get a 404 Error on trying access local file system with solid-file-client 1.0.2. Something like `fc.readFolder` or `fc.readFile` fails with `file:` calls but not with `https:`.  

I am not able to say where the error is as I cannot get the unbundled version to run because it mixes cjs and esm calls. I am also unable to run tests; the `TEST_PREFIX` thing stumps me - I have never used jest!

node: 12.16.1
os: Windows 10 x64

CxRes:
I got the tests to work!

Strangely on `file://`,  only `tests/SolidApi.links.test.js` fails. I could provide you a stack dump if you like.

On `app://` its more of a mess. On `https://` I am not sure how to setup login... 

Again Windows is more of a pain in general (but where mom and pop users are).

Otto-AA:
I've run it with travis in a windows environment now and it reproduced your errors with `file://` at `tests/SolidApi.links.test.js`.

The immediate cause of it is, that it expects relative paths, but gets absolute paths like here:
```
  ● copying links › copyAclFileForItem › modify absolute paths in acl file of a file to match new location agent:to_target

    expect(received).resolves.toEqual(expected) // deep equality

    - Expected  - 3
    + Received  + 3

      ↵
      @prefix : <#>.
      @prefix n0: <http://www.w3.org/ns/auth/acl#>.
      @prefix n1: <http://xmlns.com/foaf/0.1/>.
    - @prefix c: </profile/card#>.
    + @prefix c: <file://C:\Users\travis\build\Otto-AA\solid-file-client/profile/card#>.

      :ControlReadWrite
          a n0:Authorization;
    -     n0:accessTo <./target.txt>;
    +     n0:accessTo <file://C:\Users\travis\build\Otto-AA\solid-file-client/test-folder/SolidApi-links/copying-links/target.txt>;
          n0:accessTo <./target.txt>;
    -     n0:agent </profile/card#me>;
    +     n0:agent <file://C:\Users\travis\build\Otto-AA\solid-file-client/profile/card#me>;
          n0:agent c:me;
          n0:agentClass n1:Agent;
          n0:default <./>;
          n0:mode n0:Control, n0:Read, n0:Write.
```

So the problem is likely the `getRootUrl` method in SolidApi.links.test.js which uses `/`:
```javascript
const getRootUrl = url => {
  const base = url.split('/')
  let rootUrl = base[0]
  let j = 0
[...]

PR #144, opened 2020-04-21
Fix environment variables for tests in windows

Otto-AA:
It looks like windows doesn't support setting environment variables like `TEST_PREFIX=https:// jest`, that's why the travis build completely failed previously. I've changed it so travis sets the environment variables and the tests for windows actually run.

As noted by @CxRes , it looks like the test for windows fails for `tests/SolidApi.links.test.js` with the prefix `file://`. I haven't looked much into this issue, but I will my initial findings on these errors to #130 .


Otto-AA:
(Don't merge yet, still need to check something in travis. Doesn't look like it should...)

Otto-AA:
Nevermind, the travis setup looks good now. The test setup is just buggy for windows.

CxRes:
May i suggest that you install "cross-env" for global variables. or if you like I could set it up and push it in a day or so?

Otto-AA:
Thanks for that suggestion, it looks like a good idea. I will add it and update this PR in a few minutes.

On travis I would still use the env option, because it makes it easier to view the different builds in different environments. So I will leave that for now

PR #172, opened 2020-11-18
v1.2.0

bourgeoa:
- solid-node-client replace mock of solid-auth-cli
  this needed to add $rdf = require('redflib') as global in jest for PATCH
- extend rdf-query to write- rewrite PATCH tests to use rdf-query write
- add rdf-query tests


jeff-zucker:
Could you explain this:

> extend rdf-query to write- rewrite PATCH tests to use rdf-query write

Why are you using rdf-query rather than rdflib?

bourgeoa:
It is only in the tests and dates back from the first PATCH with N3
Because I knew how to write N-triples with N3 and N3 was there not rdflib.
At some point the turtle content produced by the tests was difficult to check.
Testing the result was much more easier.

Issue #175, opened 2020-11-24
sfc should refuse to delete POD root


scenaristeur:
move function has deleted the root of my POD 
https://github.com/solid/solidcommunity.net/issues/27

bourgeoa:
@scenaristeur v1.2.2 should also close issue for deleteFolder() which is a recursive delete.
SFC do not correct simple delete() on root or rootAcl. This is a fetch delete the correction is left to NSS.
In practice it will fail with 409 unless you delete all contents before.

Issue #176, opened 2020-11-25
sfc should refuse to copyFolder, moveFolder to a parent url with options.merge = replace


bourgeoa:
options.merge = replace in copyFolder, moveFolder delete recursively the destination folder, deleting the whole tree including the origin.

bourgeoa:
@scenaristeur v1.2.2 should close issue

PR #177, opened 2020-11-25
edit copy/moveFolder() and deleteFolderContents()

bourgeoa:
@jeff-zucker 
This PR should correct 2 bugs :
- copy/moveFolder() shall not allow `options.merge = replace` `to` a folder in the same tree as `from` 
- deleteFolderContents() of the pod root shall be forbidden

Tests added.


bourgeoa:
Travis tests fails only windows and on file://
The failing test do not appear to be related to this PR which is a security one.

The PR should close issues #175 and #176

PR #186, opened 2021-02-02
slash semantics

bourgeoa:
to createFolder(), putFile(), createFile()


jeff-zucker (changes requested):
Thanks much. I prefer this wording :  
```
Can not use createFolder to create a file.
Can not use  postFile to create a folder.
Can not use putFile to create a folder.
```

bourgeoa:
@jeff-zucker done

PR #218, opened 2022-01-23
add comment. start exploration in this file.

ewingson:
d2021-01-23 t21:59

### DIFF
--- a/docs/examples/index.html
+++ b/docs/examples/index.html
@@ -38,10 +38,11 @@ body {
     border:1px solid black;
 }
 </style>
-</head><body>
-<!-- I think this goes here -->
+<!-- it's at place here -->
     <script src="https://cdn.jsdelivr.net/npm/@inrupt/solid-client-authn-browser@1/dist/solid-client-authn.bundle.js">
 </script>
+</head><body>
+
 <div id="sidebar">
     <h3>Solid-File-Client Demo</h3>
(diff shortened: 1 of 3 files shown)


jeff-zucker:
@ewingson please work in any way that is comfortable for you.  Comments are good.  We can always clean up later.  If you need my review or have questions, please ping me in chat.

ewingson:
I'm forever thankful that finally, after all this observation, I get the opportunity to get my hands dirty. ;-)

ewingson:
the CDN has its scope in all the files of the app, hasn't it ?
I work for the beginning with ctrl-c/ctrl-v from the example.
and the philosophy of submitting little and often, is that okay ?

ewingson:
@jeff-zucker 
(I put it under the old PR cause my github branch has no issues offered by github)
could you please review https://github.com/ewingson/solid-file-client/tree/use_dpop_auth/docs/examples/advanced (both files)

several q's opened

1. are the dependencies all needed ?
2. are they in the right order ?
3. is it correct to call the actual code at the very bottom ?
4. `session` is used in `index` L77 and L103. does that matter ? 
5. another duplicate `getDefaultSession` in `index` L77 and copy.js L2 ?

statements
I wanna trace the option buttons / checkboxes / ids / classes.
I wanna use existing code wherever possible.

to really work it / control it ... I must `understand` it.
what do you think ?

PR #219, opened 2022-01-28
switch demo to new auth

ewingson:
d2022-01-28 t18:59
@bourgeoa @jeff-zucker 

the code is functional. I did adjustments.
expect this as my first contribution of actual working code.
what still needs to be done is an options menu for the pod provider.

### DIFF
--- a/docs/examples/index.html
+++ b/docs/examples/index.html
@@ -1,7 +1,7 @@
 <head>
 <!-- Bootstrap CSS
 -->
-<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous"></link>
+<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
 
 <style>
 body {
--- a/docs/examples/index.html
+++ b/docs/examples/index.html
@@ -67,4 +67,3 @@ document.getElementById('return').addEventListener( 'click', e => {
 })
 </script>
 </body>
-</html>
--- a/docs/examples/index.html
+++ b/docs/examples/index.html
@@ -66,3 +66,5 @@ document.getElementById('return').addEventListener( 'click', e => {
    document.location.href="https://github.com/jeff-zucker/solid-file-client/blob/master/README.md"
 })
 </script>
+</body>
+</html>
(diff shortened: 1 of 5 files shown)


jeff-zucker:
Great work, @ewingson, here's your first merged PR!

bourgeoa:
@ewingson great job.
If you keep the Vite dependency you must document the purpose and how to use it.
For the purpose of podProvider the simplest is just under a JavaScript `prompt` examples must be as simple as possible.

ewingson:
@bourgeoa @jeff-zucker 

a) I am afraid the `vite` dependency happened by mistake. first, I installed it during the development in a wrong folder, so that npm thought it would be needed. I am new to this.

b) I know what a prompt is, but I'm not sure I understand what you mean. would a javascript input field with a prefilled content (solidweb.me in that case) be sufficient or what do you suppose do do exactly in this example ?

bourgeoa:
b) is perfect. I was too quick

ewingson:
I am afraid I am not yet ready.
I just went to https://jeff-zucker.github.io/solid-file-client/docs/examples/ and the Login gave me an Invalid Request. that must happen because of the local version being (treated) other than the live merged version.
I suppose the following:

1. maybe we delete the `vite` dependency, case it is not technical necessary
2. I make a new fork and a new clone of the repo
3. I install `vite` in the right directory (the dir of the app, in this case `docs/examples/upload` )
4. I try to make it work locally 
5. what we have to do to make it work remote remains still unclear for me.

bourgeoa:
I went to try the app and it is not working.
- I cannot login : I suppose because it is in an Iframe. DPoP do not work in Iframe with SolidOS also
- also if you select `copy` and click on `login` you are back to the old authorization flow.

ewingson:
this comes because `docs/examples/` is the index of the app and copy/upload happens in different iframes. so work to be done. but before I fork and clone new, we should solve the dependency question

ewingson:
@jeff-zucker @bourgeoa 
(comment on location: I misplaced this edit here https://github.com/jeff-zucker/solid-file-client/pull/218#issuecomment-1026451304 cause my repo doesn't have issues. I'm working through the stuff)
could you please review https://github.com/ewingson/solid-file-client/tree/use_dpop_auth/docs/examples/advanced (both files)

several q's opened

1. are the dependencies all needed ?
2. are they in the right order ?
3. is it correct to call the actual code at the very bottom (script call L119) ?
4. `session` is used in `index.html` L77 and L103. does that matter ? 
5. another duplicate `getDefaultSession` in `index.html` L77 and `copy.js` L2 ?

statements
I wanna trace the option buttons / checkboxes / ids / classes.
I wanna use existing code wherever possible.

to really work it / control it ... I must `understand` it.
what do you think ?

ewingson:
status: upload working
copy not working
I have made copy verbose and am trying to follow the codeflow
Bildschirmfoto_2022-02-06_13-24-30

ewingson:
@bourgeoa @jeff-zucker @SharonStratsianis

I post a screenshot
Bildschirmfoto_2022-02-07_16-09-42

jeff-zucker:
I was wrong, the .acl errors you get are from solid-file-client you should trap them by putting a try/catch around the copy commands.  Leaving the console logs is ok with me.

PR #224, opened 2022-02-16
demo update error handling

ewingson:
d2022-02-16 t07:32

@bourgeoa I've set the import source to get rid of the 404. working locally. please merge to test remote.


ewingson:
@bourgeoa 

*great*, the 404 vanished. the progress bar is doing, what it's intended to do. only, we haven't got a `favicon.ico` in the root dir.

